PR #140, opened 2021-08-31
Add proper ownership references

mhoffm-aiven:
This PR aims to add a notion of ownership to aiven resources. For example we want to be able to delete an Kafka Topic when the owning Kafka is deleted.


mhoffm-aiven:
Not sure if this is the right approach, closing for now, keeping the branch for reference.

Issue #98, opened 2021-07-06
Improve API Reference docs


jonatasbaldin:
The [API Reference](https://aiven.github.io/aiven-kubernetes-operator/docs/api-reference/) docs is auto generated from the source code (comments on the Types).

We need an overall look on that and improve the comments accordingly.

We also need to add a Table of Contents.

PR #159, opened 2021-09-14
generate api docs in markdown

mhoffm-aiven:
Rewrite API Reference Generation with Markdown templates to make be able to generate a table of contents.

Solves #98 

Before:
image

After: 
image

Caveats:
* we lose nested tables ( not too bad i would argue ) since markdown does not feature them
* currently achors are broken because the doc generator library generates them with HTML in mind


ivan-savciuc (approved):
Looks good

jonatasbaldin (approved):
looks great, thanks!

PR #164, opened 2021-09-17
make directive for scorecard tests

mhoffm-aiven:


ivan-savciuc (approved):
LGTM, also works well

PR #167, opened 2021-09-22
allow deleting unprocessed instances when the API token is invalid

ivan-savciuc:


mhoffm-aiven (approved):
LGTM, very welcome change

PR #169, opened 2021-11-15
fix makefile generate and manifest commands

ivan-savciuc:
Neither of these two commands was not generating anything, fixing a path


mhoffm-aiven (approved):
lgtm, thanks

PR #172, opened 2021-11-17
fix missing kafka acl rbac

mhoffm-aiven:
This PR fixes the issue that we are not able to manage kafka ACLs right now because of missing RBACs for the status


ivan-savciuc (approved):
LGTM

PR #188, opened 2022-04-14
Enable termination protection

chazu:
This PR adds the ability for Postgres CRs to be configured for termination protection. It has been tested manually and the happy path case works: CRs which specify termination protection do indeed have termination protection, and once removed from the CR configuration, they can be deleted. 

I hesitated to add a test case out of unfamiliarity with Gomega and concerns about bloating the tests. However, I'm happy to add a test case if desired for merging.

Please let me know if there are any additional changes or enhancements which you'd like to see before this can be merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ test-acc: test-check-project-env test-check-token-env $(GINKGO) $(ENVTEST_TOOLS)
 	KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT=120s \
 	KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT=true \
 	KUBEBUILDER_ASSETS=$(abspath $(ENVTEST_TOOLS_DIR)) \
-	$(GINKGO) \
+	$(GINKGO) focus \
 		--nodes=4 \
 		--race \
 		--randomizeAllSpecs \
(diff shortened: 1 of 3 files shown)


chazu:
It appears that these test failures are CI-related - of course "it works on my machine" isn't conclusive but the error message suggests as much.

ivan-savciuc:
@chazu thanks for contributing and sorry for coming a bit late to you. This PR triggered some internal discussions about this product.

The discussion that we have internally is about web-hooks and weather we wan't to have them mandatory and configurable or not. You may wonder how this is related to this PR? We already have a K8s termination protection in the web-hooks and deliberately excluded it from the reconcile loop, the rational behind it was that when someone attempts to delete a K8s resource and termination protection is ON, web-hook will imminently report a problem without  even calling the reconciler.  

Since web-hooks are optional right now and we let this PR through a customer on attempt to delete a resource will not get any errors and reconciler will  keep trying to delete a resource for the end of times spamming logs with the errors. So if we would make web-hooks mandatory we are happy to accept this however we would need to extend this for all other services like Kafka, Opensearch, etc. 

I know that you guys having some troubles running our operator on GCP K8s cluster, and we suspect this is because the port that we are using for web-hooks is closed, and we will fix that and other reported issues in the next release. \

What is your take in general about making web-hooks mandatory?

ivan-savciuc:
> It appears that these test failures are CI-related - of course "it works on my machine" isn't conclusive but the error message suggests as much.

it looks like our CI configuration issue, code looks great, thought we would need to extend this to other CR like Kafka, Opensearch, etc.

esalter:
We understand the operator itself enforces termination protection (through the webhook) but we also wanted to enable it through the API, as sort of a fail-safe, and it also protects from accidental deletion through the UI.

We see deletion as being a very deliberate act and so if we really desire the k8s resource to be deleted and corresponding services then we would disable termination protection manually in those cases. That's the logic we were coming from anyway. We completely understand there are multiple use cases here.

chazu:
Oh, and thanks very much for your thoughtful response 🙇

ivan-savciuc (approved):
I'm merging this PR and will add Controller level termination protection for other resources shortly.

ivan-savciuc:
@chazu could you please update this branch?

PR #205, opened 2022-09-16
Allow use of default aiven token instead of authSecretReference on resources

### LINKED ISSUE #183: Ability to specify default aiven token
We would like to be able to specify a default aiven token to use, so that we don't have to distribute tokens to everyone who wants to create a new instance of anything.

Ideally, it would work so that if `authSecretRef` is not specified on the CR, use the default token configured on the operator.

This would allow us to use Kubernetes mechanisms (RBAC and OPA) to restrict what users can create, and at the same time ensure that they can't use the token to do anything else. For instance, we would let them create Redis and OpenSearch instances, but not Postgres or Kafka. If they have access to an API-token, they can do "anything".

mortenlj:
This will allow running the operator with a default token, instead of having to add a secret in every namespace where resources are created.

We had some trouble deciding on how to test this, so we made only a minimal change adding an additional test to redis_controller_test.go. This doesn't really cover the different cases properly, but we were unable to find a suitable way to do it. We are unfamiliar with Ginkgo, so might be some tricks we didn't find that would have solved it.

Since the test is mostly a simple cut&paste and minor change, we didn't want to do it for all controllers, when the actual logic change is in common code.

We tried running `make test-e2e` to get a proper check that it worked, but that didn't seem to run even before our changes.
`make test-acc` works as it should.

Fixes #183

### DIFF
--- a/api/v1alpha1/clickhouse_types.go
+++ b/api/v1alpha1/clickhouse_types.go
@@ -11,7 +11,7 @@ type ClickhouseSpec struct {
 	ServiceCommonSpec `json:",inline"`
 
 	// Authentication reference to Aiven token in a secret
-	AuthSecretRef AuthSecretReference `json:"authSecretRef"`
+	AuthSecretRef AuthSecretReference `json:"authSecretRef,omitempty"`
 
 	// Information regarding secret creation
 	ConnInfoSecretTarget ConnInfoSecretTarget `json:"connInfoSecretTarget,omitempty"`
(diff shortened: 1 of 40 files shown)


Serpentiel:
hey 👋 

thank you for your contribution! our team will evaluate your PR soon and will reach out with feedback, if any!

ivan-savciuc (approved):
@mortenlj thanks for this contribution! :+1: 

I will check what is happening with the e2e tests and release a new version where a user can set a default token via helm.

Acc tests:
```
...
Ran 17 of 17 Specs in 878.327 seconds
SUCCESS! -- 17 Passed | 0 Failed | 0 Pending | 0 Skipped
```

PR #206, opened 2022-10-13
add projectVPC reference for service types

byashimov:
Adds ResourceReferance feature. That makes possible to make certain resources to "wait" until "depended" resources are up and running.

With this PR comes ProjectVPC kind reference.


controllers/controller.go
```diff
@@ -224,23 +236,60 @@ func (i instanceReconcilerHelper) reconcileInstance(ctx context.Context, o clien
 	return ctrl.Result{}, nil
 }
 
-func (i instanceReconcilerHelper) checkPreconditions(o client.Object) (bool, ctrl.Result, error) {
-	i.rec.Event(o, corev1.EventTypeNormal, eventWaitingforPreconditions, "waiting for preconditions of the instance")
+func (i instanceReconcilerHelper) checkPreconditions(ctx context.Context, o client.Object) (bool, error) {
```
ivan-savciuc, line 239: What is the reason why we change this function signature? In this new implementation it handles error and re-queue  separately, so that a user cannot do both at once.

controllers/controller.go
```diff
@@ -224,23 +236,60 @@ func (i instanceReconcilerHelper) reconcileInstance(ctx context.Context, o clien
...
+	i.rec.Event(o, corev1.EventTypeNormal, eventWaitingForPreconditions, "waiting for preconditions of the instance")
+
+	// Checks references
+	if oRef, ok := o.(refsObject); ok {
+		i.log.Info("checking resource references")
+		ready, err := i.checkRefs(ctx, oRef)
+		if err != nil {
+			i.log.Info(fmt.Sprintf("one or more references can't be found yet: %s", err))
+			return true, err
```
ivan-savciuc, line 248: It doe's not make sense here since reconciliation will error and never re-queue. We need to keep the original function signature to allow this. After many errors, the k8s control plane puts resource reconciliation on halt for some significant time, so it is not a good idea to make an error without a critical reason.

ivan-savciuc: By simply fixing this code we will get:
```
--- PASS: kuttl (480.41s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/projectvpc-ref (425.37s)
```
So boost the e2e test perfonca from 21 minute to 8

ivan-savciuc (changes requested):
A first read through comments: 

This PR has 6 commits and some of them are from the same nature, please squash all of them to a one or two with the meaningful messages. 

I did run e2e tests: 
```
--- FAIL: kuttl (1297.15s)
    --- FAIL: kuttl/harness (0.00s)
        --- FAIL: kuttl/harness/connect-kafka-to-postgres (0.00s)
        --- PASS: kuttl/harness/clickhouse-simple-cluster (109.91s)
        --- PASS: kuttl/harness/redis-simple-cluster (150.56s)
        --- PASS: kuttl/harness/opensearch-simple-cluster (150.57s)
        --- PASS: kuttl/harness/kafka-simple-cluster (152.39s)
        --- PASS: kuttl/harness/simple-kafka-topic (212.24s)
        --- PASS: kuttl/harness/projectvpc-ref (1242.01s)
FAIL
```

It fails and seems unrelated to the new code, but I will run it again sometime later. What is concerning is the time it takes for the `projectvpc-ref` takes 20+ minutes alone, we need to check what is going on there.

controllers/controller.go
byashimov: Nice catch! This is definitely should not return true.

controllers/controller.go
byashimov: Because of code duplication. Now there is two more returns with requeue now. 
It's also never checked if any error returned:

```
	if requeue, result, err := i.checkPreconditions(o); requeue {
		return result, err
	}

	if !isAlreadyProcessed(o) {
```

It either returns error or reschedules reconciliation.

controllers/controller.go
byashimov: Nice catch, thanks, should return nil.

controllers/clickhouse_controller.go
```diff
@@ -39,15 +40,15 @@ func (r *ClickhouseReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func (h ClickhouseHandler) createOrUpdate(avn *aiven.Client, i client.Object) error {
+func (h ClickhouseHandler) createOrUpdate(ctx context.Context, k8s client.Client, avn *aiven.Client, i client.Object) error {
```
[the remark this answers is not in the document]
byashimov: This is possibly a good idea, but eventually handlers manage resources, and it really depends if that resource has or not any references to deal with.

controllers/controller.go
ivan-savciuc: It was intentional not to check error, why - check this one https://github.com/aiven/aiven-operator/pull/206#discussion_r999436704

ivan-savciuc (changes requested):
I think we also need to update the ProjectVPC deletion; Aiven would not allow us to delete ProjectVPC until all services that are using it are still up and running, so we should delete all the services and only then can remove ProjectVPC.

controllers/controller.go
byashimov: That is a bit confusing. It's more common to check error for nil, rather than ignoring. Because when you extend checkPreconditions function then you have to know that it breaks the pattern. 
For instance you can send both requeue=false and err != nil and then it will pass. Error should not be ignored.

ivan-savciuc (changes requested):
- Please add to the service webhooks validation to make it impossible to set a `projectVPCRef` and `projectVpcId` simultaneously. 
- Add to `ProjectVPC.delete` method graceful error handing of `VPC cannot be deleted while there are services migrating from it 409`

ivan-savciuc:
Also found a bug, when you do something like that: 
```
kubectl apply -f test/e2e/projectvpc/projectvpc-ref/01-resources.yaml ;kubectl delete -f test/e2e/projectvpc/projectvpc-ref/01-resources.yaml
projectvpc.aiven.io/k8s-e2e-projectvpc-ref created
kafka.aiven.io/k8s-e2e-projectvpc-ref-kafka created
opensearch.aiven.io/k8s-e2e-projectvpc-ref-os created
redis.aiven.io/k8s-e2e-projectvpc-ref-redis created
projectvpc.aiven.io "k8s-e2e-projectvpc-ref" deleted
kafka.aiven.io "k8s-e2e-projectvpc-ref-kafka" deleted
opensearch.aiven.io "k8s-e2e-projectvpc-ref-os" deleted
redis.aiven.io "k8s-e2e-projectvpc-ref-redis" deleted
```

`ProjectVPC` resource delete function runs forever because it relies on the `Status.ID`, and if we try to delete `ProjectVPC` fast, like in the example from above `status.ID` never gets set.

controllers/controller.go
```diff
@@ -224,23 +236,59 @@ func (i instanceReconcilerHelper) reconcileInstance(ctx context.Context, o clien
...
+func (i instanceReconcilerHelper) checkRefs(ctx context.Context, o refsObject) (bool, error) {
+	for _, r := range o.GetRefs() {
+		i.log.WithValues("ResourceReference", r).Info("processing reference")
+		obj, err := v1alpha1.RefToObject(ctx, i.k8s, r)
+		if err != nil {
+			return false, err
+		}
 
-	return false, ctrl.Result{}, nil
+		if !(isAlreadyProcessed(obj) && isAlreadyRunning(obj)) {
```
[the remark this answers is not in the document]
byashimov: If I get it right we need both to be true:

```
if not processed or not running:
    return not ready (false)
```

can be turned to into:
```
if not (processed and running):
```

not (A and B) = (not A) or (not B),  https://en.wikipedia.org/wiki/De_Morgan%27s_laws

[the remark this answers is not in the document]
ivan-savciuc: All right, I haven't seen the brackets `(...)` before not `!`

ivan-savciuc (approved):
LGTM

PR #207, opened 2022-10-14
update kubernetes operator with the new operator sdk layout

ivan-savciuc:
This PR includes the following changes:
- Fixes the `testenv` installation and RBAC issues on the  Kubernetes >= 1.2.
- Upgrade the project with the latest v3 operator SDK layout and make it more compatible for further upgrades.
- Simplify the Makefile and make it in line with the newest Operator SDK template
- Fix and rework e2e tests

Known limitations:
- acc tests are using ginkgo `v1` and now run sequentially even if we set ginkgo to run in parallel with N nodes
- release process might require some adjustments
- all CI jobs need to be checked 

TODOs:
- now we can write webhook tests
- try to upgrade `ginkgo` to v2 and run acc tests in parallel
- enable `prometheus` monitoring for operator
- propagate new metrics, prometheus and default token via helm


Serpentiel (approved):
lgtm

byashimov (approved):

PR #219, opened 2023-01-04
add userconfig generator, add MySQL kind, add genericServiceHandler

byashimov:
- Adds user config generator (`go:generate`)
- Adds `genericServiceHandler` to generalize service management 
- Adds MySQL Kind
- Fixes acceptance tests


Makefile
```diff
@@ -109,6 +117,7 @@ test: manifests generate fmt vet envtest ginkgo ## Run tests.
 	KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT=true \
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" \
 	$(GINKGO) \
+			--output-interceptor-mode=none \
```
mhoffm-aiven, line 120: indentation seems weird

main.go
```diff
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
+	aiveniov1alpha1 "github.com/aiven/aiven-operator/api/v1alpha1"
```
mhoffm-aiven, line 19: duplicate import

controllers/generic_service_handler.go
```diff
@@ -0,0 +1,149 @@
...
+	if !exists && !aiven.IsNotFound(err) {
+		return fmt.Errorf("failed to fetch service: %w", err)
+	}
+
+	// Creates if not exists or updates existing service
+	var reason string
+	if !exists {
+		reason = "Created"
+		req := o.newCreateRequest()
+		req.ProjectVPCID = toOptionalStringPointer(projectVPCID)
```
Serpentiel, line 54: to deduplicate this part a bit, create a `projectVPCIDPtr` variable before `if !exists` and use it in the block instead:

```suggestion
		req.ProjectVPCID = projectVPCIDPtr
```

controllers/generic_service_handler.go
```diff
@@ -0,0 +1,149 @@
...
+			return err
+		}
+		_, err = a.Services.Create(spec.Project, req)
+		if err != nil {
+			return fmt.Errorf("failed to create service: %w", err)
+		}
+	} else {
+		reason = "Updated"
+		req := o.newUpdateRequest()
+		req.ProjectVPCID = toOptionalStringPointer(projectVPCID)
```
Serpentiel, line 66: and then you can reuse `projectVPCIDPtr` in this conditional branch too

```suggestion
		req.ProjectVPCID = projectVPCIDPtr
```

userconfigs_generator/generator_test_source.json
```diff
@@ -0,0 +1,869 @@
+{
```
Serpentiel, line 1: since we have switched to `aiven-go-client/tools/exp` for the source, shouldn't this JSON also be changed to be YAML and make use of the new schema?

userconfigs_generator/generator_test_expected.txt
```diff
@@ -0,0 +1,484 @@
+// Code generated by user config generator. DO NOT EDIT.
```
Serpentiel, line 1: since this isn't really a text file, could the file be renamed to have `.go` extension in it? you could also consider something like `.tmp.go`, or `.unused.go`, etc.

userconfigs_generator/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"strings"
+
+	"github.com/aiven/aiven-go-client/tools/exp/dist"
+)
+
+const destination = "./api/v1alpha1/userconfigs"
+
+func main() {
+	var serviceList string
+	flag.StringVar(&serviceList, "services", "", "Comma separated service list of names to generate for")
```
Serpentiel, line 15: could we please use https://github.com/spf13/cobra to manage CLIness of this tool, as this is the preferred way of introducing CLI tools within PRs to Go projects of Aiven organization?

`cobra` is a drop-in replacement for built-in `flag` package, and it shouldn't be too hard to switch to it here, please see https://github.com/aiven/aiven-go-client/blob/master/tools/exp/cmd/root.go for example usage

Serpentiel (changes requested):

userconfigs_generator/generator_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewUserConfigFile(t *testing.T) {
+	src, err := os.ReadFile(`generator_test_source.json`)
+	assert.NoError(t, err)
+
+	obj := new(object)
+	err = json.Unmarshal(src, obj)
```
Serpentiel, line 16: I believe this should also use the `gopkg.in/yaml.v3` package

Serpentiel (changes requested):

userconfigs_generator/main.go
byashimov: Cobra is an overkill for a just one flag and an unnecessary dependancy with indirect dependencies.

controllers/generic_service_handler.go
byashimov: I believe that doesn't make any difference.

controllers/generic_service_handler.go
byashimov: I believe that doesn't make any difference. It is still a duplicated assignment.

Serpentiel: this is easier to break as it adds an extra branch—if someone wants to change the way this functions, they can easily forget about the other line, so please change it

controllers/generic_service_handler.go
Serpentiel: see above thread

userconfigs_generator/main.go
Serpentiel: well, that was exactly what happened to my PR before as well, and I believe there might be a specific reason why we would want to use Cobra

perhaps @ivan-savciuc could share some additional info on it, since they asked me previously to add Cobra when I were introducing a simple CLI tool with one flag

Serpentiel (dismissed):
please check conversations again

main.go
```diff
@@ -30,6 +32,7 @@ const port = 9443
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
+	utilruntime.Must(v1alpha1.AddToScheme(scheme))
```
ivan-savciuc, line 35: This line is a dublicate, please remove.

userconfigs_generator/main.go
ivan-savciuc: I do not see a point of using Cobra framework here.

controllers/mysql_controller.go
```diff
@@ -0,0 +1,130 @@
...
+func newMySQLAdapter(object client.Object) (serviceAdapter, error) {
+	mysql, ok := object.(*v1alpha1.MySQL)
+	if !ok {
+		return nil, fmt.Errorf("object is not of type v1alpha1.MySQL")
+	}
+	return &mySQLAdapter{mysql}, nil
+}
+
+// mySQLAdapter handles an Aiven MySQL service
+type mySQLAdapter struct {
```
ivan-savciuc, line 48: Something for the future, perhaps all the adapter functions accept `newSecret` can be generalized further to avoid a copy&paste when we expand this logic to other services.

main.go
ivan-savciuc, line 35: This line is a duplicate, please remove.

ivan-savciuc (changes requested):
Please remove this duplicate

controllers/generic_service_handler.go
byashimov: See my above answer.

controllers/generic_service_handler.go
byashimov: See my above answer.

ivan-savciuc (approved):
LGTM

PR #220, opened 2023-01-12
fix aiven client useragent

### LINKED ISSUE #196: User Agent does not include version number
In internal logs we are seeing the K8s Operator appear with a `User-Agent` header as follows:

```
"k8s-operator/"
```

It looks like the version number is not correctly appended in the two places where we call `aiven.NewTokenClient()`.

I think it would also be good to make the UA a little more Aiven specific by appending `-aiven` to it like we do in the Terraform Provider, ie `"k8s-operator-aiven"` because then we can know with more confidence where these API calls are coming from.

byashimov:
Proxy client version to the user agent. Resolves #196


Serpentiel (approved):
lgtm

PR #228, opened 2023-01-18
fix kafka acl lifecycle

byashimov:
Recreates Kafka ACL if it's been modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 - Add Cassandra Kind
 - Add Grafana Kind
+- Recreate Kafka ACL if modified. 
+  Note: Modification of ACL created prior to v0.5.1 won't delete existing instance at Aiven.
+  It must be deleted manually.
 
 ## v0.6.0 - 2023-01-16
 
--- a/controllers/kafkaacl_controller.go
+++ b/controllers/kafkaacl_controller.go
@@ -44,26 +44,29 @@ func (h KafkaACLHandler) createOrUpdate(avn *aiven.Client, i client.Object, refs
 		return err
 	}
 
-	exists, err := h.exists(avn, acl)
+	// ACL can't be really modified
+	// Tries to delete it instead
+	_, err = h.delete(avn, i)
 	if err != nil {
 		return err
 	}
 
-	if !exists {
-		_, err = avn.KafkaACLs.Create(
-			acl.Spec.Project,
-			acl.Spec.ServiceName,
-			aiven.CreateKafkaACLRequest{
-				Permission: acl.Spec.Permission,
-				Topic:      acl.Spec.Topic,
-				Username:   acl.Spec.Username,
-			},
-		)
-		if err != nil && !aiven.IsAlreadyExists(err) {
-			return err
-		}
+	// Creates it from scratch
+	r, err := avn.KafkaACLs.Create(
+		acl.Spec.Project,
+		acl.Spec.ServiceName,
+		aiven.CreateKafkaACLRequest{
+			Permission: acl.Spec.Permission,
+			Topic:      acl.Spec.Topic,
+			Username:   acl.Spec.Username,
+		},
+	)
+	if err != nil {
+		return err
 	}
 
+	// New created ACL id set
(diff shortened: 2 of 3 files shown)


Serpentiel (dismissed):
lgtm

palmieric (dismissed):
lgtm!

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 - Add Cassandra Kind
 - Add Grafana Kind
+- Recreate Kafka ACL if modified.
```
ivan-savciuc, line 6: Is this actually true that we need to ask users to delete ACL? Since we delete the old ACL before creating a new one.

controllers/kafkaacl_controller.go
```diff
@@ -140,12 +131,16 @@ func (h KafkaACLHandler) get(avn *aiven.Client, i client.Object) (*corev1.Secret
...
 
-	aivenACL, err := h.getAivenACL(avn, acl)
+	id, err := h.getID(avn, acl)
 	if err != nil {
 		return nil, err
 	}
 
-	acl.Status.ID = aivenACL.ID
+	// todo: remove in v1
+	if acl.Status.ID == "" {
```
ivan-savciuc, line 140: Why do we need this if? Can we just have `acl.Status.ID = id`?

controllers/kafkaacl_controller_test.go
```diff
@@ -71,11 +71,42 @@ var _ = Describe("Kafka ACL Controller", func() {
...
 			Expect(meta.IsStatusConditionTrue(createdACL.Status.Conditions, conditionTypeRunning)).Should(BeTrue())
+
+			By("checking initial permission before update")
+			Expect(createdACL.Spec.Permission).Should(Equal("admin"))
+
+			By(`updating permission from "admin" to "write" and making sure it has a new ID`)
+			createdACL.Spec.Permission = "write"
+			Expect(k8sClient.Update(ctx, createdACL)).Should(Succeed())
+
+			// It takes a time
```
ivan-savciuc, line 86: Typo here, should be `It takes time` without a

controllers/kafkaacl_controller_test.go
```diff
@@ -71,11 +71,42 @@ var _ = Describe("Kafka ACL Controller", func() {
...
+			var updatedACL *v1alpha1.KafkaACL
+			Eventually(func() bool {
+				updatedACL = &v1alpha1.KafkaACL{}
```
ivan-savciuc, line 89: It is not a good idea to query Aiven during the acc tests. In these tests, we are interested in observing K8s CR state transitions as our customers would. For example, once we altered the spec, a new generation was created, and the resource got reconciled and went through different conditions. From the customer's perspective, what should they do to get an updated ACL from their K8s application while waiting for that change?

ivan-savciuc (changes requested):
Just a few small comments, otherwise LGTM.

CHANGELOG.md
byashimov: Old ACLs doesn't have id. They are found in the `List()` by topic name, username and permission. If you modify the manifest you can't find them, hence delete it. That is the thing, yes (

controllers/kafkaacl_controller.go
byashimov: Just want make it explicit.

controllers/kafkaacl_controller_test.go
byashimov: It queries k8s :)

Makefile
```diff
@@ -131,7 +131,7 @@ build: generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: manifests generate install fmt vet ## Run a controller from your host.
-	go run ./main.go
+	ENABLE_WEBHOOKS=false go run ./main.go
```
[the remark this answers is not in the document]
byashimov: Better remove it then. It outputs to much no one will ever notice. Better have it in my zshrc.

controllers/kafkaacl_controller.go
byashimov: Replaced the whole block btw.

ivan-savciuc (approved):
LGTM

PR #242, opened 2023-01-25
ci(tests): separate job per each package

byashimov:
Runs tests separately.

### DIFF
--- a/.github/workflows/acceptance-tests.yaml
+++ b/.github/workflows/acceptance-tests.yaml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  acceptance_tests:
+  test:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


byashimov:
Workflow https://github.com/aiven/aiven-operator/actions/runs/4008201172

.github/workflows/acceptance-tests.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  acceptance_tests:
+  test:
```
Serpentiel, line 7: please don't rename it, it will create a duplicate in the GH Actions history, which I would need to clean manually again because there's no button to delete all the entries from it

Serpentiel (changes requested):
lgtm, but one small change

Serpentiel (approved):
lgtm

PR #247, opened 2023-01-29
fix: projectvpc deletion

ivan-savciuc:
ProjectVPC resource was undeletable by the webhook that had the update operation disabled. And our reconciler couldn't remove a finalized from the Project VPC resource. So the solution is to remove the ProjectVPC webhook entirely (since the update block was the only logic) and mark all fields as immutable.


byashimov:
Please checkout [this file](https://github.com/aiven/aiven-operator/blob/720b42a4801607859c286e380b1019e9b1b20a3d/config/crd/kustomization.yaml).
It has some patches.

Serpentiel (approved):
lgtm

PR #248, opened 2023-01-29
fix: unable to remove resources with invalid token

ivan-savciuc:
When a resource was successfully created, and after that, a token expires or becomes invalid, error handling is supposed to pass such cases and successfully finalize the resource and remove it from k8s.


Serpentiel (approved):
lgtm

PR #251, opened 2023-01-31
fix docs static generation, downgrade docsy

byashimov:
- use docsy compatible version
- fix docs static generation


Serpentiel (approved):
lgtm

PR #253, opened 2023-02-01
add service integrations

byashimov:
Adds `read_replica` service integration.


byashimov:
tests https://github.com/aiven/aiven-operator/actions/runs/4085370494

ivan-savciuc:
@byashimov pls rebase

controllers/generic_service_handler.go
```diff
@@ -69,6 +72,25 @@ func (h *genericServiceHandler) createOrUpdate(a *aiven.Client, object client.Ob
 			UserConfig:            userConfig,
 		}
 
+		for _, s := range spec.ServiceIntegrations {
+			// Validates that read_replica is running
+			// If not, the wrapper controller will try later
+			if s.IntegrationType == "read_replica" {
```
ivan-savciuc, line 78: I reckon this IF statement should be moved to the `func (h *genericServiceHandler) checkPreconditions` and in this case we do not need additional error handling in the `controller.go`

controllers/controller.go
```diff
@@ -209,7 +209,13 @@ func (i instanceReconcilerHelper) reconcileInstance(ctx context.Context, o clien
 
 	if !isAlreadyProcessed(o) {
 		i.rec.Event(o, corev1.EventTypeNormal, eventCreateOrUpdatedAtAiven, "about to create instance at aiven")
-		if err := i.createOrUpdateInstance(o, refs); err != nil {
+		err := i.createOrUpdateInstance(o, refs)
```
ivan-savciuc, line 212: As mentioned before, we can avoid this additional error handling here if we move the `read_replica` service integration check to the `checkPreconditions` method of the service. 

We already have logic in our controller handler responsible for checking all possible dependencies we need before we start the creation or update of the service.

ivan-savciuc (changes requested):
A few small suggestions, otherwise LGTM. I haven't tested it thought..

controllers/generic_service_handler.go
ivan-savciuc: And also we have this helper function `checkServiceIsRunning`

ivan-savciuc (approved):
LGTM

PR #257, opened 2023-02-07
upgrade opensearch kind user config and controller

byashimov:
- Upgrade OpenSearch user config
- Use generic handler for OpenSearch type


byashimov:
tests https://github.com/aiven/aiven-operator/actions/runs/4112211614

PR #258, opened 2023-02-07
simplify service adapter

byashimov:
- Simplify service adapter: move common parts to CommonSpec
- Upgrade KafkaConnect


byashimov:
DiskSpace is not supported by KafkaConnect, so it can't be moved to CommonSpecs.

PR #265, opened 2023-02-13
Building arm64 image

### LINKED ISSUE #264: Build for linux/arm64 (multi arch)
We're using the ARM based instances in AWS (graviton), and we can't run the aiven operator on those machines as you're only building for `linux/amd64`. 

It would be great if you could start building multi arch images. It won't require a huge effort on you end, just a few changes to the Dockerfile and the GH actions workflow.

https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

christoffer-eide:
Fixes #264

Note that I have not tested the GH actions workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -43,6 +43,9 @@ jobs:
 
     - id: action_vars
       run: echo ::set-output name=TAG::${GITHUB_REF#refs/tags/}
+    - uses: docker/setup-qemu-action@v1
+      with:
+        platforms: arm64
     - uses: docker/setup-buildx-action@v2
     - uses: docker/login-action@v2
       with:
@@ -52,4 +55,5 @@ jobs:
       uses: docker/build-push-action@v2
       with:
         push: true
+        platforms: linux/amd64,linux/arm64
         tags: aivenoy/aiven-operator:${{ steps.action_vars.outputs.TAG }}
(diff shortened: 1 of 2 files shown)


Serpentiel:
hey 👋 

thanks for your contribution! I will check the PR now

.github/workflows/release.yaml
```diff
@@ -43,6 +43,9 @@ jobs:
 
     - id: action_vars
       run: echo ::set-output name=TAG::${GITHUB_REF#refs/tags/}
+    - uses: docker/setup-qemu-action@v1
+      with:
+        platforms: arm64
```
Serpentiel, line 48: doesn't this breaks `amd64` build then?

Serpentiel (changes requested):

.github/workflows/release.yaml
christoffer-eide: Don't think so as the GH actions runners run on `amd64`. I could add `amd64` just to be safe? 
The whole GH actions build would need to be tested by you guys

christoffer-eide: I'm not 100% sure if quemu is needed in the first place

Serpentiel: ahh, looks like this whole `with` block [is not needed](https://www.winters.nz/docker/cross-platform-github-actions/), could you please remove it?

Serpentiel: thanks, we'll see if it works or not :)

Serpentiel (approved):
lgtm

PR #267, opened 2023-02-14
fix(userconfig): string enums parsed as int

byashimov:
YAML validator validates string enums with integer values only as integers. For instance, postgres version `14` is validated as int and can't pass the validation. That makes impossible to set values to that kind of fields.


byashimov:
tests https://github.com/aiven/aiven-operator/actions/runs/4174774860

Serpentiel (approved):
lgtm

PR #271, opened 2023-02-17
refactor(userconfigs): split service and integration

byashimov:
KafkaConnect exists in both service and integrations config. This request separates configs for future changes, and doesn't change CRDs


Serpentiel (approved):
lgtm

PR #276, opened 2023-02-20
make auth object fields required

byashimov:
Marks auth field required


ivan-savciuc (changes requested):
Not sure if it is a good idea, our users have two possibility's to set an Aiven secret:
- Via AuthSecretRef
- Via global token
And the logic is the following: if an global token is set, we use it, but if not, we check for the authSecretRef.

So, this change will block the global token logic

ivan-savciuc (changes requested):
Since we are making authRef a pointer, I think we also would need to change logic here https://github.com/aiven/aiven-operator/blob/b279b48e2757ee93ffb04bc6589d9bd46b39e3fb/controllers/controller.go#L118 and perhaps in other places as well

byashimov:
> Since we are making authRef a pointer, I think we also would need to change logic here
> 
> https://github.com/aiven/aiven-operator/blob/b279b48e2757ee93ffb04bc6589d9bd46b39e3fb/controllers/controller.go#L118
> 
> and perhaps in other places as well

There is an [initial check](https://github.com/aiven/aiven-operator/pull/276/files#diff-74ab0909653e09e5a3dd2d7af63242d429528a2a6434e564372a72e49e3b9195R76), so it should be fine.
Add another one right to the basic controller.

byashimov:
tests are green https://github.com/aiven/aiven-operator/actions/runs/4231808709

ivan-savciuc (approved):
LGTM

PR #282, opened 2023-03-01
[bug]Setting config: min_cleanable_dirty_ratio: 0.1 does not apply

TV2rd:
Hello, I bumped aiven-operator to 0.8.0 in our test cluster and tried to set: 

`apiVersion: aiven.io/v1alpha1
kind: KafkaTopic
metadata:
  name: aiven-kafka-topic
  namespace: kafka-sandbox
spec:
  authSecretRef:
    name: aiven-token
    key: token
  config:
    min_cleanable_dirty_ratio: 0.2
  termination_protection: false
  project: some-aiven-project
  serviceName: aiven-kafka-sandbox

  partitions: 1
  replication: 2⏎`
Aiven-Operator accepted the manifest but min_cleanable_dirty_ratio set was not reflected in Aiven console. There it was  set to 0.5 with source: "Kafka Default Configuration". 
I checked the commit and the MinCleanableDirtyRatio was missing from convertKafkaTopicConfig used in create/update func. Also if the tests is a full integration test running against some Aiven service using 0.5 as the test value would lead to a false negative result. 

Cheers
Tord


byashimov:
Hi. Thanks for this!

byashimov (approved):

Issue #209, opened 2022-11-05
Change mapping of KafkaTopic.spec.termination_protection to *bool


christoffer-eide:
We're trying to make some [kyverno](https://kyverno.io/docs/) polices to enforce/validate kafka topics created using the operator.

One of the issues we have run into, is related to validation of `KafkaTopic.spec.termination_protection`. We want to enforce that this field is set to either `true` or `false` when creating the topic. The goal is to "force" the devs to decide if they want termination protection or not when the topic is initially created.

Testing with the `v0.4.0` release, it's not possible to to differentiate between `termination_protection: false` and the field being unset/missing.

This is caused by the boolean [field mapping](https://github.com/aiven/aiven-operator/blob/v0.4.0/api/v1alpha1/kafkatopic_types.go#L38):
```
type KafkaTopicSpec struct {
  [...]
  TerminationProtection bool `json:"termination_protection,omitempty"`
}
```
If `TerminationProtection` could use a `*bool` pointer instead, it would be possible to [distinguish between `true`, `false` and unset/missing](
https://github.com/golang/go/issues/13284#issuecomment-158153815).

PR #285, opened 2023-03-02
feat: accept false value for termination_protection

byashimov:
Type bool with `json:omitempty` doesn't accept `false` value.
Resolves: #209


ivan-savciuc (approved):
LGTM

PR #307, opened 2023-03-20
feat(helm): copy helm charts from external repository

byashimov:
Adds helm charts.

### DIFF
--- a/charts/aiven-operator-crds/.helmignore
+++ b/charts/aiven-operator-crds/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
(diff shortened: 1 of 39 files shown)


mhoffm-aiven:
The other repository was also set up as a helm repository via github actions, do we do this here too?

byashimov:
> The other repository was also set up as a helm repository via github actions, do we do this here too?

Yes. The idea to automate charts releasing process (patching). I have already made a "script" for that. But I do want to have clean diff rather than putting altogether in one huge messy PR.

So this, and [this](https://github.com/aiven/aiven-operator/pull/308) are preparations for a real feature.

mhoffm-aiven:
Ok great! I intially did set it up as external repo since i was thinking that we might have other helm charts too ( for `karapace` i.e. ) but if we only have the operator it makes more sense to have it grouped here!

byashimov:
> Ok great! I intially did set it up as external repo since i was thinking that we might have other helm charts too ( for `karapace` i.e. ) but if we only have the operator it makes more sense to have it grouped here!

Yeah, that would make sense.
Just want to reduce risks on copy-pasting things. And eventually I want to release thing altogether automatically using _version_ number, rather than putting that manually everywhere.

Serpentiel (approved):
lgtm

PR #310, opened 2023-03-21
refactor(ip_filter): remove backward compatibility

byashimov:
Ip filter objects introduced in v0.8.0.
Since we had some breaking changes there, this compatibility layer doesn't help at all.


Serpentiel (approved):
lgtm

PR #311, opened 2023-03-21
fix(delete): deletion errors omitted

byashimov:
Some errors, like termination protection, has been omitted.
This fix creates warning event with the omitted error.

### DIFF
--- a/controllers/basic_controller.go
+++ b/controllers/basic_controller.go
@@ -85,6 +85,7 @@ const (
 	eventTryingToDeleteAtAiven              = "TryingToDeleteAtAiven"
 	eventUnableToDeleteAtAiven              = "UnableToDeleteAtAiven"
 	eventUnableToDeleteFinalizer            = "UnableToDeleteFinalizer"
+	eventUnableToDelete                     = "UnableToDelete"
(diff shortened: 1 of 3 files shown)


Serpentiel (approved):
lgtm

controllers/basic_controller.go
```diff
@@ -337,6 +338,9 @@ func (i instanceReconcilerHelper) finalize(ctx context.Context, o client.Object)
 		} else if aiven.IsNotFound(err) {
 			i.rec.Event(o, corev1.EventTypeWarning, eventUnableToDeleteAtAiven, err.Error())
 			return ctrl.Result{}, fmt.Errorf("unable to delete instance at aiven: %w", err)
+		} else {
```
ivan-savciuc, line 341: It may make more sense to refactor this using a switch case, WDYT.

ivan-savciuc (approved):
LGTM

controllers/basic_controller.go
byashimov: Makes sense, but git blame then will show my changes instead of original PR with feature/fix. So I better keep it that way :)

PR #316, opened 2023-03-24
refactor(tests): test postgresql with the new suite

byashimov:
- Adds postgresql test on the new suite
- Removes gingko test and generic handler test, both cases are handled with the new test


Serpentiel (dismissed):
lgtm

Serpentiel (approved):

PR #332, opened 2023-04-04
feat: add rbac options to helm chart

ryaneorth:
This change allows the Helm chart be installed without including the creation of cluster-wide resources (ClusterRoles/ClusterRoleBindings). Specifically, this change supports at least 2 use cases:

1. The scenario where the engineer installing the chart has limited (i.e. namespace only) access and cannot install cluster-wide resources. This change allows the cluster-wide resources to be installed separately by cluster administrators while still allowing the engineers to manage the operator itself without having full cluster-wide access.

2. The scenario where the operator only needs to manage resources in a specific namespace and not the entire cluster.

I believe this is the first pull request into this repository since moving the Aiven charts from its [dedicated repository](https://github.com/aiven/aiven-charts) so if a separate PR needs to be made there, let me know and I'd be happy to open one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add Kafka `log_cleaner_min_cleanable_ratio` minimum and maximum validation rules
 - Remove Kafka version `3.2`, reached EOL
 - Remove PostgreSQL version `10`, reached EOL
+- Add `rbac.create` and `rbac.scoped` options to Helm chart
 
 ## v0.9.0 - 2023-03-03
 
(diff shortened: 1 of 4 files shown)


byashimov:
Hi @ryaneorth !
Thanks for the contribution! May I ask you to add a line to the changelog, please?
Regarding moving the charts to this rep, it is still in progress. I need to do few things first that may take a week or two. Then we can release this charts.
Is this urgent for you (the new chart)? Just to know if I need to prioritise this. Thanks.

byashimov (changes requested):
Please update the changelog.

ryaneorth:
Hi @byashimov - I've added a line item to the change log. This isn't incredibly urgent, but if we could have a new version of the chart released with these changes in the next week or so that would be very helpful!

byashimov (approved):

PR #346, opened 2023-04-11
feat(projectvpc): explicitly delete by id, speed up deletion

byashimov:
- Deletes `ProjectVPC` by id
- Speed ups deletion by accepting `DELETED` status as deleted


Serpentiel (approved):
lgtm

PR #358, opened 2023-04-13
revert(charts): add rbac options to helm chart

byashimov:
Reverts #332

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@
 - Add Kafka `log_cleaner_min_cleanable_ratio` minimum and maximum validation rules
 - Remove Kafka version `3.2`, reached EOL
 - Remove PostgreSQL version `10`, reached EOL
-- Add `rbac.create` and `rbac.scoped` options to Helm chart. Thanks to @ryaneorth
 - Explicitly delete `ProjectVPC` by `ID` to avoid conflicts 
 - Speed up `ProjectVPC` deletion by exiting on `DELETING` status
 - Fix missing RBAC permissions to update finalizers for various controllers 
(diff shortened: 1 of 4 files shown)


Serpentiel (approved):

ryaneorth:
Hey @byashimov - out of curiosity why was this removed?

byashimov:
@ryaneorth hey!
We were just about to release the new version when this broke the operator. So it was reverted asap. I should have left this comment at least, but I was busy by the fix, sorry.
Our operator is designed to watch [all namespaces](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#manager-watching-options), that's why it requires cluster wide role.
So to make this feature happen, we need to pass a list of namespaces to listen to helm values, and then to the operator as env variables.
Said that, it is not enough to change the role only.
If you still need this feature, I can create an issue on github. You can do it also if you have more to add. Then we can take this to the backlog :)

ryaneorth:
hey @byashimov ! My particular use-case was actually [number 1 outlined in the original PR](https://github.com/aiven/aiven-operator/pull/332#issue-1654079513). Specifically to support:

> The scenario where the engineer installing the chart has limited (i.e. namespace only) access and cannot install cluster-wide resources. This change allows the cluster-wide resources to be installed separately by cluster administrators while still allowing the engineers to manage the operator itself without having full cluster-wide access.

Given this, would it be possible to re-introduce the part of change that supports this use-case but not the second use-case (which I don't need) where you can optionally install non-cluster-wide roles/role bindings?

byashimov:
hey @ryaneorth!
Currently operator will request permissions to listen the whole cluster. So if you set limited role it won't work (well, it doesn't in our clusters, or probably we miss something?).
We have this in the backlog now. But I'm not sure I can say when this one is released :(

PR #368, opened 2023-04-18
ci: fix github actions yaml spec errors

Serpentiel:
fixes GitHub Actions YAML spec errors, also sets Go version to use `go.mod` everywhere

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -1,26 +1,23 @@
 name: pages
 
 on:
-  workflow_dispatch:
+  workflow_dispatch: {}
   push:
     tags:
-      - 'v*'
+      - v*
 
 jobs:
   release_pages:
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-
       - name: Generate docs
         uses: addnab/docker-run-action@v3
         with:
           image: squidfunk/mkdocs-material:latest
           options: -v ${{ github.workspace }}/docs/:/docs
           run: mkdocs build
-
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,57 +3,54 @@ name: release
 on:
   push:
     tags:
-      - 'v*'
+      - v*
 
 jobs:
   build_default_release_manifest:
     runs-on: ubuntu-latest
     steps:
-
-    - uses: actions/checkout@v3
-    - id: action_vars
-      run: echo ::set-output name=TAG::${GITHUB_REF#refs/tags/}
-    - shell: bash
-      run: |
-        echo "resources: [manager.yaml]
-        apiVersion: kustomize.config.k8s.io/v1beta1
-        kind: Kustomization
-        images:
-        - name: controller
-          newName: aivenoy/aiven-operator
-          newTag: ${{ steps.action_vars.outputs.TAG }}
-        " > config/manager/kustomization.yaml
-    - uses: karancode/kustomize-github-action@master
-      with:
-        kustomize_version: 4.3.0
-        kustomize_build_dir: config/default
-        kustomize_output_file: deployment.yaml
-      env:
-        GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_ACCESS_TOKEN }}
-    - uses: svenstaro/upload-release-action@v2
-      with:
-        repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: deployment.yaml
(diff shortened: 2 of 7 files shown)


.github/workflows/acceptance-tests.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.18'
+          go-version: stable
```
byashimov, line 13: I haven't tested 1.20. Please bring back 1.19

.github/workflows/pages.yaml
```diff
@@ -1,10 +1,10 @@
 name: pages
 
 on:
-  workflow_dispatch:
+  workflow_dispatch: {}
```
byashimov, line 4: Empty object, why?

.github/workflows/release.yaml
```diff
@@ -54,6 +54,6 @@ jobs:
       with:
         push: true
         platforms: linux/amd64,linux/arm64
-        tags: aivenoy/aiven-operator:${{ steps.action_vars.outputs.TAG }}
+        tags: aivenoy/aiven-operator:${{ steps.action_vars.outputs.tag }}
         build-args: |
-          VERSION=${{steps.action_vars.outputs.TAG}}
+          VERSION=${{ steps.action_vars.outputs.tag }}
```
byashimov, line 59: Please have tested this.

.github/workflows/tests.yaml
```diff
@@ -1,7 +1,7 @@
 name: tests
 
 on:
-  workflow_dispatch:
+  - workflow_dispatch
```
byashimov, line 4: Let's have `on` of the same style across all workflows. It is a list here and an object in `pages`. Worth to mention, it is an object in [docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow).

.github/workflows/tests.yaml
```diff
@@ -13,10 +13,10 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.19'
+          go-version: stable
```
byashimov, line 16: Please revert.

byashimov (changes requested):

.github/workflows/pages.yaml
Serpentiel: YAML spec

.github/workflows/release.yaml
Serpentiel: it changes the name

.github/workflows/tests.yaml
Serpentiel: GitHub's official linter suggests this to be an array when the object functionality is not used

.github/workflows/tests.yaml
Serpentiel: see first response

.github/workflows/tests.yaml
byashimov: Then please change it for pages.

> Let's have on of the same style across all workflows. It is a list here and an object in pages.

.github/workflows/acceptance-tests.yaml
byashimov: Then what it does? [Documentation](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file) says you should use different option then. And for `go-version`:

> go-version: '1.18.0-rc.1' # The Go version to download (if necessary) and use.
> If stable is provided, action will get the latest stable version from the [go-versions](https://github.com/actions/go-versions/blob/main/versions-manifest.json) repository manifest.

Serpentiel: I have provided an answer to about this during our internal discussion, but here's a follow-up either way: changing the `go-version` from `1.18` to `stable` in this workflow doesn't actually change the version of Go being used—instead, it sets the Go version to be the latest `stable` release, while still respecting the [Go 1 compatibility promise](https://go.dev/doc/go1compat)—this means that any valid Go program written under Go 1.x should continue to work without modification under any future Go 1.y release; additionally, packages and programs compiled with earlier versions of Go will continue to work with future versions of the Go compiler as the Go project has made a commitment to maintain the stability of the language and its standard library, so developers can rely on their code continuing to work as expected even as new versions of Go are released

Serpentiel: changed to `go-version-file: go.mod`

.github/workflows/tests.yaml
Serpentiel: I've standardized this across all the workflows 🎉

byashimov (approved):
lgtm

PR #376, opened 2023-04-21
feat(secrets): add annotations and label fields

### LINKED ISSUE #195: Add annotations and labels to generated secrets
In our organization, different teams have different namespaces, but we want them to be able to use the same Kafka cluster. The plan is to use SystemUsers for this purpose. This will generate secrets in the same namespace as the ServiceUser, where the aiven-token secret is accessible. In order to distribute the generated secret to other namespaces, something like [reflector](https://github.com/emberstack/kubernetes-reflector) can be used. But this uses annotations to allow secrets to be copied to other namespaces. The only way to do this now is to run something like this, after the secret has been generated:
```
kubectl annotate secret kafka-crab-connection reflector.v1.k8s.emberstack.com/reflection-allowed=true
```
A better way would be to be able to specify the annotation i want directly on the ServiceUser resource. I can imagine this functionality can be useful for labels too.

Example ServiceUser with annotations/labels:
```
apiVersion: aiven.io/v1alpha1
kind: ServiceUser
metadata:
  # the name of our user 🦀
  name: crab
spec:
  authSecretRef:
    name: aiven-token
    key: token

  # the Secret name we will store the users' connection information
  # looks exactly the same as the Secret generated when creating the Kafka cluster
  # we will use this Secret to produce and consume events later!
  connInfoSecretTarget:
    name: kafka-crab-connection
    annotations:
      reflector.v1.k8s.emberstack.com/reflection-allowed: "true"
    labels:
      team: "red"

  # the Aiven project the user is related to
  project: my-project

  # the name of our Kafka Service
  serviceName: my-kafka-service

```

byashimov:
Resolves #195


byashimov:
tests are [running](https://github.com/aiven/aiven-operator/actions/runs/4765038971/jobs/8470317649)

ivan-savciuc (commented):
Is this the first part of this feature implementation? 

I assume we need to propagate these to the secret properties Aiven Operator will create for a service, WDYT?

ivan-savciuc (commented):
In the end we should have something like: 
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-secret-1
  annotations:
    key1: "val1"
   ...
 labels:
   l1: "label1"
  ..
spec:
 ...
```

byashimov:
That's correct. Was checking the idea should have marked it draft.

byashimov:
[Tests](https://github.com/aiven/aiven-operator/actions/runs/4768756911/jobs/8478443346)

Serpentiel:
good job on adding this, really good stuff 👍

I'll pass on reviewing this, let's wait for @ivan-savciuc to check this out

ivan-savciuc (approved):
LGTM

ivan-savciuc:
At some point would be great to extend our tests with some metadata

PR #381, opened 2023-04-25
feat: add option to add prefix on each connInfoSecretTarget secrets' keys

jordiclariana:
This PR adds the option to connInfoSecretTarget to add prefixes to all its secret keys.

This intends to help to avoid collision environment variables when the secrets created by the operator are used in any `env[].valueFrom.secretKeyRef`. This way there are easily identified as coming from the operator and the application/pod consuming them does not have to adjust to variable name collisions.

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -32,6 +32,8 @@ type ConnInfoSecretTarget struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// Labels added to the secret
 	Labels map[string]string `json:"labels,omitempty"`
+	// Prefix added to all secret's keys
+	Prefix string `json:"prefix,omitempty"`
 }
 
 // ServiceStatus defines the observed state of service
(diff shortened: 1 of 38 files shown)


byashimov:
Hey @jordiclariana!
Thanks for the contribution. There will be some refactoring around the secrets. I know this problem exists, some resources have same secret keys. I want to fix that, but just replacing them is a major breaking change.
I thought to set a default value per kind, so it can be overridden softly. Like for `Pg` kind set default prefix `PG_`, then have secrets `PG_HOST`, `PG_PORT`, etc. This approach will let existing clusters to have a backward compatible solution, and let us fix this collision issue. 
Your solution is quite close to that.
I'll talk to the colleagues, and come back with the final decision.

byashimov:
Hey @jordiclariana!
Does [this](https://github.com/aiven/aiven-operator/pull/391) solution work for you?

jordiclariana:
I replied in the PR https://github.com/aiven/aiven-operator/pull/391#issuecomment-1536160436

byashimov:
Hey. Thank you for the contribution. I'm closing this PR as we have merged #391

PR #384, opened 2023-04-26
feat: update service tags when creating or updating a service

### LINKED ISSUE #383: Tags are not working
When creating a service with tags, the service is created without the specified tags.

Seems the implementation is not completed.

mortenlj:
Fixes #383

### DIFF
--- a/tests/postgresql_test.go
+++ b/tests/postgresql_test.go
@@ -95,9 +95,11 @@ func TestPgReadReplica(t *testing.T) {
 	assert.Equal(t, masterAvn.State, master.Status.State)
 	assert.Equal(t, masterAvn.Plan, master.Spec.Plan)
 	assert.Equal(t, masterAvn.CloudName, master.Spec.CloudName)
-	assert.Equal(t, map[string]string{"env": "prod", "instance": "master"}, master.Spec.Tags)
 	assert.NotNil(t, masterAvn.UserConfig) // "Aiven instance has defaults set"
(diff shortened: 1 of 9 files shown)


byashimov:
Hey!
Thanks for the contribution. Nice catch.
Could you please restore tags comparison line? Because now it says two maps are equal, but they might be both empty as well.

mortenlj:
> Could you please restore tags comparison line? Because now it says two maps are equal, but they might be both empty as well.

Fair point, fixed now.

byashimov (approved):
lgtm, thanks for the contribution!

tests/postgresql_test.go
```diff
@@ -106,7 +109,10 @@ func TestPgReadReplica(t *testing.T) {
 	assert.Equal(t, replicaAvn.State, replica.Status.State)
 	assert.Equal(t, replicaAvn.Plan, replica.Spec.Plan)
 	assert.Equal(t, replicaAvn.CloudName, replica.Spec.CloudName)
-	assert.Equal(t, map[string]string{"env": "test", "instance": "replica"}, replica.Spec.Tags)
+	assert.Equal(t, map[string]string{"env": "test", "instance": "foo"}, replica.Spec.Tags)
```
byashimov, line 112: Hmmm. I was too hasty.

mortenlj: Oh, oops. I did that change with a search-replace, guess I didn't check the diff well enough.

But I think the test passed ... 🤔

PR #387, opened 2023-05-02
fix(tags): service tags update

byashimov:
Fixes #384


tests/suite.go
```diff
@@ -260,6 +275,10 @@ func isNotFound(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "not found")
 }
 
+func alreadyExists(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "already exists")
```
Serpentiel, line 279: perhaps `errors.Is` would work better here?

Serpentiel (approved):
lgtm

tests/suite.go
byashimov: It comes with custom object type. So once it is mixed into a string you can't determine it by internal function.

```
func IsAlreadyExists(err error) bool {
	return ReasonForError(err) == metav1.StatusReasonAlreadyExists
}
```

PR #391, opened 2023-05-04
feat(secrets): use kind name as default secret's prefix

byashimov:
Use kind name as prefix for secrets to avoid collisions.
Old secrets will be removed in future releases.

### DIFF
--- a/controllers/cassandra_controller.go
+++ b/controllers/cassandra_controller.go
@@ -68,15 +68,15 @@ func (a *cassandraAdapter) getUserConfig() any {
 
 func (a *cassandraAdapter) newSecret(s *aiven.Service) (*corev1.Secret, error) {
 	stringData := map[string]string{
-		"CASSANDRA_HOST":     s.URIParams["host"],
-		"CASSANDRA_PORT":     s.URIParams["port"],
-		"CASSANDRA_USER":     s.URIParams["user"],
-		"CASSANDRA_PASSWORD": s.URIParams["password"],
-		"CASSANDRA_URI":      s.URI,
-		"CASSANDRA_HOSTS":    strings.Join(s.ConnectionInfo.CassandraHosts, ","),
+		"HOST":     s.URIParams["host"],
+		"PORT":     s.URIParams["port"],
+		"USER":     s.URIParams["user"],
+		"PASSWORD": s.URIParams["password"],
+		"URI":      s.URI,
+		"HOSTS":    strings.Join(s.ConnectionInfo.CassandraHosts, ","),
 	}
 
-	return newSecret(a, a.Spec.ConnInfoSecretTarget, stringData), nil
+	return newSecret(a, stringData, true), nil
 }
 
 func (a *cassandraAdapter) getServiceType() string {
--- a/controllers/mysql_controller.go
+++ b/controllers/mysql_controller.go
@@ -67,17 +67,17 @@ func (a *mySQLAdapter) getUserConfig() any {
 
 func (a *mySQLAdapter) newSecret(s *aiven.Service) (*corev1.Secret, error) {
 	stringData := map[string]string{
-		"MYSQL_HOST":        s.URIParams["host"],
-		"MYSQL_PORT":        s.URIParams["port"],
-		"MYSQL_DATABASE":    s.URIParams["dbname"],
-		"MYSQL_USER":        s.URIParams["user"],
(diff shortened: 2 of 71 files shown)


byashimov:
Hey @mortenlj, @jordiclariana!
Would this solution work for you?

jordiclariana:
Hi @byashimov!

My honest opinion is that this is an improvement, but it does not solve all the cases where a collision can happen. For instance, an application might need 2 redis instances (1 for cache and the other one for sessions). With this implementation, you cannot do that with aiven-operator, since the environment variables mounted into the deployment/pod, will overlap each other. Also, the new prefix in the env vars is good, but it does not avoid possible collisions with already existing env vars (it makes it more difficult, but not impossible, and most importantly, it does not provide a method to avoid it). Hence my proposal [here](https://github.com/aiven/aiven-operator/pull/381) that you already saw (allow me to link it here for visibility), which in my opinion gave way more flexibility.

In any case, thanks for the prompt response, really appreciated. And if you reconsider my implementation, maybe we can refine it to make it more aligned with your coding guidelines if necessary.

jordiclariana:
Also, this will be a breaking change in coming releases, which can be quite annoying with large setups.

mortenlj:
We have decided to pivot slightly, so this is not important for us (we handle the secrets in our own operator), but I do think @jordiclariana has a point with how this would work when an application needs two Redis instances for instance.

byashimov:
Hi @jordiclariana
Thank you for your detailed response. 
Nice catch with two redises case.

So we have several issues here:
1. var's name collision between services of different type
2. var's name collision between services of the same type
3. backward compatibility
4. and for postgres: some names have prefix, some do not, so it can't be solved with a custom prefix:
   ```go
   "PGSSLMODE":    s.URIParams["sslmode"],
   "DATABASE_URI": s.URI,
   ```

I've merged both solutions into this PR,
and left credits for the prefix feature to @jordiclariana in the changelog.

It should solve all the cases now.
Does it work for you?

byashimov:
Tests are [green](https://github.com/aiven/aiven-operator/actions/runs/4913828721).

jordiclariana:
It definitely works for me! Thanks so much for the heads up and the work done.

controllers/cassandra_controller.go
```diff
@@ -68,15 +68,15 @@ func (a *cassandraAdapter) getUserConfig() any {
 
 func (a *cassandraAdapter) newSecret(s *aiven.Service) (*corev1.Secret, error) {
 	stringData := map[string]string{
-		"CASSANDRA_HOST":     s.URIParams["host"],
-		"CASSANDRA_PORT":     s.URIParams["port"],
-		"CASSANDRA_USER":     s.URIParams["user"],
-		"CASSANDRA_PASSWORD": s.URIParams["password"],
-		"CASSANDRA_URI":      s.URI,
-		"CASSANDRA_HOSTS":    strings.Join(s.ConnectionInfo.CassandraHosts, ","),
+		"HOST":     s.URIParams["host"],
```
ivan-savciuc, line 71: should we also add here a prefix?

controllers/mysql_controller.go
```diff
@@ -67,17 +67,17 @@ func (a *mySQLAdapter) getUserConfig() any {
...
 	stringData := map[string]string{
-		"MYSQL_HOST":        s.URIParams["host"],
-		"MYSQL_PORT":        s.URIParams["port"],
-		"MYSQL_DATABASE":    s.URIParams["dbname"],
-		"MYSQL_USER":        s.URIParams["user"],
-		"MYSQL_PASSWORD":    s.URIParams["password"],
-		"MYSQL_SSL_MODE":    s.URIParams["ssl-mode"],
-		"MYSQL_URI":         s.URI,
-		"MYSQL_REPLICA_URI": s.ConnectionInfo.MySQLReplicaURI,
+		"HOST":        s.URIParams["host"],
```
ivan-savciuc, line 70: The same here, do we need a prefix here as well?

ivan-savciuc (commented):
I think we have preix missing is some places, can you check?

controllers/cassandra_controller.go
byashimov: Prefix added automatically in newSecret func 🙂

ivan-savciuc (approved):
LGTM

PR #393, opened 2023-05-05
feat: allow optional installation of cluster role

ryaneorth:
This change allows the Helm chart be installed without including the creation of cluster-wide resources (ClusterRoles/ClusterRoleBindings). Specifically, this change supports the use case where the engineer installing the chart has limited (i.e. namespace only) access and cannot install cluster-wide resources. This change allows the cluster-wide resources to be installed separately by cluster administrators while still allowing the engineers to manage the operator itself without having full cluster-wide access.

It is understood that if `clusterRole.create=false` is set without the cluster role/cluster role binding being installed beforehand the operator will not function as expected but this gives the option to have the namespaced resources installed/managed by individuals with namepace-only access without giving access to create cluster-wide resources.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Add `annotations` and `labels` fields to `connInfoSecretTarget`
 - Allow to disable capabilities check to install webhooks. Thanks to @amstee
 - Set `OpenSearch.spec.userConfig.opensearch.search_max_buckets` maximum to `65536`
+- Add `clusterRole.create` option to Helm chart
 
 ## v0.10.0 - 2023-04-17
 
--- a/charts/aiven-operator/templates/cluster_role.yaml
+++ b/charts/aiven-operator/templates/cluster_role.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.clusterRole.create -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -602,3 +603,4 @@ rules:
       - get
       - list
       - update
+{{- end }}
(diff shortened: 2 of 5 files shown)


byashimov:
Hey. Just one question before we merge it. Does it work for you as expected, have you tested this?

ryaneorth:
Thanks for taking the time to take a look! Yes, this works as expected. The role definitions are the same as what you have in `charts/aiven-operator/templates/cluster_role.yaml` and `charts/aiven-operator/templates/cluster_role_binding.yaml` but would be installed via a separate step by a cluster administrator before this chart is installed by an engineer with namespace-only access.

I've also updated the docs indicating how others can make use of this setting.

ivan-savciuc:
@ryaneorth thanks for you contribution :+1: 

I see one problem with that solution, operator will still try to listen for all namespaces regardless of the RBAC settings. 
As you can see our `Manager` has no `Namespaces` restriction: https://github.com/aiven/aiven-operator/blob/d6fe2e8dc764dfdaff24916469331dd75f949df0/main.go#L58-L65

```go
        // Namespace, if specified, restricts the manager's cache to watch objects in
	// the desired namespace. Defaults to all namespaces.
	//
	// Note: If a namespace is specified, controllers can still Watch for a
	// cluster-scoped resource (e.g Node). For namespaced resources, the cache
	// will only hold objects from the desired namespace.
	Namespace string
```

I would propose adding `NewCache:  cache.MultiNamespacedCacheBuilder(namespaces),` to the manager, more info on this here:  https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder
And we can propagate `namespaces` as an environment variable (or a flag). If it is empty, `Manager` will be watching resources in ALL namespaces (the default behaviour we have now). In your use case, it will be watching resources only in a set of namespaces that we can configure using Helm chart. 

@ryaneorth WDYT? 

Moreover we can extend Helm charts with the logic to add RBAC only to the set of namespaces we want instead of the cluster wide role.

ivan-savciuc (approved):
LGTM

PR #413, opened 2023-05-22
fix: make conditions and state optional attributes of service status

mortenlj:
In some cases, the someone might attempt to pass an object with a `status` attribute, but with no `conditions` or `state`.
When attempting to read such objects you may encounter errors about the missing attributes.
When marking them with `omitempty`, the generated schema will no longer consider these attributes required, which probably is more correct, and allows parsers to correctly parse these objects.


ivan-savciuc (approved):
LGTM

PR #436, opened 2023-06-20
fix: finalize non-running service with expired token

byashimov:
Fixes resource deletion even with expired or missing token.


ivan-savciuc (approved):
LGTM

Issue #522, opened 2023-10-23
Error when updating ServiceIntegration for prometheus integration


mortenlj:
Creating a service integration of type `prometheus` works fine, the integration is connected and seems to be working.
If I attempt to make a change, the operator fails with an error:

```
ERROR	Reconciler error	{"controller": "serviceintegration", "controllerGroup": "aiven.io", "controllerKind":"ServiceIntegration", "ServiceIntegration": {"name":"opensearch-basseng-documents-prometheus","namespace":"basseng"}, "namespace": "basseng", "name": "opensearch-basseng-documents-prometheus", "reconcileID": "75c85631-6ba4-4190-9142-b4137150b502", "error": "unable to create or update instance at aiven: unable to create or update aiven instance: 400: {\"errors\":[{\"message\":\"Invalid input for user_config 'Service type-specific settings': None is not of type 'object'\",\"status\":400}],\"message\":\"Invalid input for user_config 'Service type-specific settings':None is not of type 'object'\"} - "}
```

From what I can gather, this is because prometheus integrations doesn't have any `user_config`, and the operator sends a nil value when calling the API. The API apparently requires that the `user_config` is an object.
Strangely enough, the API accepts a nil value when creating the service integration, but not when updating it, which might indicate that this could be considered an API bug.

PR #526, opened 2023-10-27
fix(service_intergration): empty user config to API convertion

ivan-savciuc:


byashimov (approved):
nit: switch instead of Contains call will do the job too, but overall lgtm

byashimov:
Resolves #522

PR #534, opened 2023-11-10
feat(docs): add string format

### LINKED ISSUE #528: disk_space field on OpenSearch resources is poorly documented and validated
https://aiven.github.io/aiven-operator/api-reference/opensearch.html#spec.disk_space-property

When creating a new resource of type OpenSearch with a `disk_space` property set, the newly created OpenSearch service has only the default amount of disk for the selected plan.

Changing the property after creation triggers a reconciliation, but no change happens in the service.

byashimov:
Renders string format next to a string field in docs. Resolves #528


Serpentiel (approved):
lgtm

PR #548, opened 2023-11-21
refactor(models): rename receivers to in

byashimov:
SDK generator generates `in` receiver name. While our code is inconsistent it is hard to apply changes to models using simple "find and replace" command. The PR renames receivers to `in`.


Serpentiel (approved):
lgtm

PR #549, opened 2023-11-21
feat(models): add conditions method to aiven managed object

byashimov:
This is a pre-refactoring PR.

- Extends AivenManagedObject with Conditions() method
- Moves AivenManagedObject to models


Serpentiel (approved):
lgtm

PR #550, opened 2023-11-21
fix(controllers): object not committed properly

byashimov:
This PR:
- fixes object updates lose during reconciliation
- sets error conditions on basic controller level

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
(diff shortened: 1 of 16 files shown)


tkaesserfm:
@byashimov yes, you were right! :laughing:

byashimov:
> @byashimov yes, you were right! 😆

Hey. What do you mean? :)

tkaesserfm:
@byashimov yes, but i only see the issue for KafkaACL. 
We're only using kafka-related stuff though. What works is ServiceUser and KafkaTopic.

tkaesserfm:
Ah i see, i was talking "Status Conditions" so different thing but the symptom seems the same.

ivan-savciuc (approved):
LGTM

PR #555, opened 2023-11-29
docs: link to prereqs from installation instructions

staceysalamon-aiven:
Add a link to the prerequisites list to both sets of installation instructions to make sure customers are aware of what they need before installing.


byashimov (approved):
lgtm

PR #561, opened 2023-12-11
feat(connection_pool): expose pool name

### LINKED ISSUE #560: connectionpool.connInfoSecretTarget does not expose CONNECTIONPOOL_NAME
It is not possible to get the name of the connectionpool from connInfoSecretTarget. The exposed vars in the secrets are:
```
DATABASE_URI
PGDATABASE
PGHOST
PGPASSWORD
PGPORT
PGSSLMODE
PGUSER
```

`DATABASE_URI` does include the connection pool name, but the value as a whole is not easy to consume for an application via ENVs in our environment.

Is it possible to add the name of the connection pool to the resulting secret ?

byashimov:
- Exposes `CONNECTIONPOOL_NAME`. Resolves #560 
- Fixes `CONNECTIONPOOL_PORT` has invalid port value (the one from the service)

### DIFF
--- a/controllers/connectionpool_controller.go
+++ b/controllers/connectionpool_controller.go
@@ -5,6 +5,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"strconv"
 
 	"github.com/aiven/aiven-go-client/v2"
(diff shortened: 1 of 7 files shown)


ivan-savciuc (approved):
LGTM

controllers/connectionpool_controller.go
```diff
@@ -170,16 +177,17 @@ func (h ConnectionPoolHandler) get(ctx context.Context, avn *aiven.Client, obj c
 
 	prefix := getSecretPrefix(connPool)
 	stringData := map[string]string{
+		prefix + "NAME":         connPool.Name,
 		prefix + "HOST":         s.URIParams["host"],
-		prefix + "PORT":         s.URIParams["port"],
```
ivan-savciuc, line 174: I have only one question - what was the problem passing PORT as it arrives from the `URIParams["port"]`? Now we pars the URI `url.Parse(cp.ConnectionURI)` and get it from there

byashimov: > I have only one question - what was the problem passing PORT as it arrives from the `URIParams["port"]`? Now we pars the URI `url.Parse(cp.ConnectionURI)` and get it from there

Well. Service has its own port. And the pool has its own one. You create a pool, you want to use its port. Currently you create a pool and still use the one that comes with the service. So not really using the pool.

byashimov: I'll add a comment there.

ivan-savciuc: Got it thanks for the clarification

PR #576, opened 2023-12-19
feat(service): add service tech email support

rriski:
Add support for service tech emails


api/v1alpha1/common.go
```diff
@@ -87,6 +88,10 @@ type ServiceCommonSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	// Service integrations to specify when creating a service. Not applied after initial service creation
 	ServiceIntegrations []*ServiceIntegrationItem `json:"serviceIntegrations,omitempty"`
+
+	// +kubebuilder:validation:MaxItems=10
+	// Technical contact emails of the service
+	TechnicalEmails *[]avnClient.ContactEmail `json:"tech_emails,omitempty"`
```
byashimov, line 94: Hey. Please don't use client structs here. We need more control on that fields. Like adding validation rules.

byashimov (changes requested):
Please add a test by extending any existing service test. Like Kafka, pg, etc.

byashimov:
Btw please make sure that existing list can be purged. It is sufficient to test the behaviour on a local cluster, because the test might be complicated a bit.

api/v1alpha1/common.go
rriski: Added test in https://github.com/aiven/aiven-operator/pull/576/commits/f94a58472144580ed311e3cc6af0df71beb28689

rriski:
> Btw please make sure that existing list can be purged. It is sufficient to test the behaviour on a local cluster, because the test might be complicated a bit.

Test case added in PR handles checking that emails can be deleted

byashimov (approved):
Great job.

PR #578, opened 2023-12-21
ci(lint): add trunk linting setup

rriski:
Best reviewed commit by commit.

* run linting with [trunk-io](https://docs.trunk.io/check)
* add markdownlint check
* add codespell check
* add yamllint check
* add hadolint check

Most of the diff is autofixed.


.github/workflows/lint.yml
```diff
@@ -9,25 +9,17 @@ on:
       - main
 
 jobs:
-  commitlint:
+  trunk:
+    if: >
```
Serpentiel, line 13: maybe we can drop this `if` check for now, we don't really have the `skip workflows` label at the moment, if this is needed, we can always add it in future

Serpentiel (changes requested):

Serpentiel:
otherwise lgtm

Serpentiel:
also please consider a different naming for the first commit/PR, `feat` makes sense in new product features most of the time, there are other good tags that work here, e.g. `ci`

rriski:
> also please consider a different naming for the first commit/PR, `feat` makes sense in new product features most of the time, there are other good tags that work here, e.g. `ci`, `build`, etc.

Fixed the first commit message and removed `if` from github action

.github/workflows/lint.yml
```diff
@@ -9,25 +9,14 @@ on:
       - main
 
 jobs:
-  commitlint:
```
Serpentiel, line 12: let's keep `commitlint`: sadly Trunk doesn't/can't have this functionality, because it has to be executed on a higher level than Trunk operates on, i.e. it should be GitHub-aware, and it's not

Serpentiel (changes requested):
lgtm but cant merge because of `commitlint`

Serpentiel (approved):
lgtm

Issue #541, opened 2023-11-16
A way to disable creation of connInfoSecret


mortenlj:
We do our own Just-In-Time provisioning of serviceusers and credentials for our Aiven services, so we don't need or use the secret created by the Aiven Operator (the "connInfoSecret").

Every now and then, some enterprising developer notices that there exists a secret with the same name as the service instance and attempts to use the credentials stored in the connInfoSecret. We would rather that they didn't, and it would be useful if we could just disable the generation of the connInfoSecret entirely.

This is somewhat related to #472 , but while that suggests the operator should use an already existing secret, we want there to be no secret at all.

byashimov:
Hey, thanks for the idea.
I'm taking this to the sprint.

PR #599, opened 2024-01-23
feat: add ability to disable connection info secret creation

rriski:
Adds a `connInfoSecretTargetDisabled` field to resources that generate connection info secrets.

The following restrictions apply:

- `connInfoSecretTargetDisabled` can only be set on newly created resources
- `connInfoSecretTargetDisabled` cannot be mutated once set

These restrictions are in place due to the following factors because we don't want to delete secrets of previously created resources that may be depended on. For the same reason, we don't allow mutating the field once set so as to not delete secrets. In theory, we could allow changing the value from `true` to `false`, which would have the effect of 1) not creating the secret at resource provisioning time because `connInfoSecretTargetDisabled: true` and 2) at update time, the secret would be created. We decided to not implement this behaviour at this time.

Resolves https://github.com/aiven/aiven-operator/issues/541


rriski:
[tests](https://github.com/aiven/aiven-operator/actions/runs/7656550410/job/20865136165)

tests/service_opts_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// THEN
+	// Check if service connection info secret is created
+	secret, err := s.GetSecret(grafana.GetName())
+	if scenario.shouldSecretBeCreated {
+		assert.NoError(t, err)
+		assert.NotEmpty(t, secret)
+	} else {
+		assert.Nil(t, secret)
+	}
```
byashimov, line 136: How about:
```
secret, err := s.GetSecret(grafana.GetName())
assert.Equal(t, scenario.expectErrIsNil, err == nil)
assert.Equal(t, scenario.expectSecretIsNil, secret == nil)
```

tests/service_opts_test.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func TestServiceConnInfoSecretTargetDisabled(t *testing.T) {
+	cases := []TestScenario{
+		{
+			// Modifying `connInfoSecretTargetDisabled` from missing to true --> A secret is created, but the update fails
+			// because `connInfoSecretTargetDisabled` can only be set during service creation.
+			name:                                "modify from missing to true",
+			initialConnInfoSecretTargetDisabled: nil,
+			updatedConnInfoSecretTargetDisabled: pointer.Bool(true),
```
byashimov, line 163: nit: Probably generating yaml in-place might make it easier to read and debug (less if/else, pointers)
```
baseYaml := getConnInfoSecretTargetDisabledYaml(project, name, cloudName)
disabledTrue := fmt.Sprintf("%s\n  connInfoSecretTargetDisabled: true", baseYaml)
disabledFalse := fmt.Sprintf("%s\n  connInfoSecretTargetDisabled: false", baseYaml)

...
{
name: "no disabled",
yaml: baseYaml,
},

{
name: "with disabled true",
yaml: disabledTrue,
},
{
name: "with disabled false",
yaml: disabledFalse,
},
```

byashimov (changes requested):
Only docstrings are mandatory to fix. Otherwise, looks great to me.

tests/service_opts_test.go
rriski: This becomes a bit difficult because the name will conflict. I agree that generating yaml in place is better. I have updated the PR with another variant, let me know if this is better.

byashimov (approved):
lgtm

PR #610, opened 2024-01-26
feat(changelog): override entries with same titles

byashimov:
- Merges entries with same changes for the same field. For instance, when enum property is updated several times during the same release
- Removes changelog duplicates


CHANGELOG.md
```diff
@@ -32,10 +32,8 @@
 - Change `KafkaConnect` field `userConfig.additional_backup_regions`: deprecated
 - Change `Kafka` field `userConfig.additional_backup_regions`: deprecated
 - Change `OpenSearch` field `userConfig.additional_backup_regions`: deprecated
-- Change `PostgreSQL` field `userConfig.pg_version`: enum ~~`[11, 12, 13, 14, 15]`~~ → `[11, 12, 13,
-14, 15, 16]`
+  14, 15, 16]`
```
rriski, line 35: This seems incorrect: 
image

byashimov: Thanks. I regenerated from changelog since the last release.

PR #622, opened 2024-02-05
ci(codeql): add setup Go step

Serpentiel:
adds setup Go step to codeql workflow


rriski (approved):
LGTM, would be good to link to where you noticed that this is needed (Github actions log or the repo security page)

Serpentiel:
@rriski I moved this from Terraform repo, I can't find it anywhere, but I think it won't hurt if we have it

PR #635, opened 2024-02-14
fix(tests): move defaultNamespace const to fix undefined variable issue

jeff-held-aiven:
This addresses an issue that has been failing any commit that occurred after #632


Serpentiel (approved):
lgtm

PR #636, opened 2024-02-14
feat(tests): log operation status when calling `retryForever`

jeff-held-aiven:
To better understand what exactly failed, `retryForever` now requires that a string that contains a description of the operation be passed.


byashimov (approved):
lgtm

PR #639, opened 2024-02-14
ci(trunk): fix trunk fmt

byashimov:
`trunk fmt` damages markdown files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,6 @@ generate: userconfigs boilerplate imports manifests docs charts fmt ## Run all c
 .PHONY: fmt
 fmt: ## Format code.
 	go fmt ./...
-	trunk fmt
 
 .PHONY: vet
 vet: ## Run go vet against code.
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -31,7 +31,7 @@ lint:
         - markdownlint
       paths:
         - .trunk/**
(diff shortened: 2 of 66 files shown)


Makefile
```diff
@@ -68,7 +68,6 @@ generate: userconfigs boilerplate imports manifests docs charts fmt ## Run all c
 .PHONY: fmt
 fmt: ## Format code.
 	go fmt ./...
-	trunk fmt
```
Serpentiel, line 71: please change the configuration in `.trunk` instead to exclude `docs` from `markdownlint`:

```yaml
lint:
  enabled:
...
    - markdownlint...
...
  ignore:
    - linters:
        - markdownlint
      paths:
        - docs/**
```

Serpentiel (changes requested):

Serpentiel:
I think this needs to be changed too to fix the YAML problem: https://github.com/aiven/aiven-operator/pull/639/files#diff-45a5a35dd7ad9a52d0fd3540f6c0d16a763575de51cc9a0a6c6f6a9f1da62ecdR29

Serpentiel:
looks like we also need to exclude `prettier` for `docs/**`

also, maybe we should keep it like it was `docs/docs/**`, because the problem was actually that `prettier` was not disabled there

Serpentiel:
solved in #643

PR #643, opened 2024-02-15
ci(trunk): fix docs linting

rriski:
* Remove trailing whitespace and blank line before EOF from docs generation template
* Don't run prettier against the docs
* Remove unused function
* Update go mods after Go 1.22 revert


.trunk/trunk.yaml
```diff
@@ -22,9 +22,10 @@ lint:
     - prettier@3.2.5
     - yamllint@1.34.0
   ignore:
-    - linters:
-        - yamllint
-        - prettier
+    - linters: [prettier]
+      paths:
+        - docs/docs/**
+    - linters: [yamllint, prettier]
```
Serpentiel, line 28: why not:

```yaml
- linters:
    - yamllint
    - prettier
```

Serpentiel (approved):
lgtm, one comment though

Serpentiel:
@rriski feel free to merge, but please reply/action on the comment

.trunk/trunk.yaml
rriski: I think the [x, y, z] format is more readable in this case. Trunk also suggests that in their docs: https://docs.trunk.io/check/configuration/ignoring-issues#ignoring-multiple-files

PR #645, opened 2024-02-16
feat: check PG upgrade task before attempting to upgrade service

### LINKED ISSUE #609: Unable to upgrade to new major version of Postgres
We attempted to upgrade to a new major version of postgres (14->15), by editing the `.spec.user_config.pg_version` field.
The change was accepted, but nothing happened in the console.
Checking the logs of the aiven-operator revealed that a missing pre-condition caused the upgrade to fail:

```
aiven-operator-7bc759484f-4g6ch aiven-operator 1.7062750644760764e+09	ERROR	Reconciler error	{"controller": "postgresql", "controllerGroup": "aiven.io", "controllerKind": "PostgreSQL", "PostgreSQL": {"name":"postgresql-basseng-evaluering","namespace":"basseng"}, "namespace": "basseng", "name": "postgresql-basseng-evaluering", "reconcileID": "2d9331d0-2bcc-4968-8863-f59dc53ab8f5", "error": "unable to create or update instance at aiven: unable to create or update aiven instance: failed to update service: 400: {\"errors\":[{\"message\":\"Upgrade check result missing, please run upgrade_check task\",\"status\":400}],\"message\":\"Upgrade check result missing, please run upgrade_check task\"} - "}
```

A bit more readable:
```json
{
  "message": "Upgrade check result missing,please run upgrade_check task",
  "status": 400
}
```

Clicking the "Upgrade version" button in the console and selecting the target version triggered an upgrade check.
When the check was completed, the operator continued with the upgrade, while we could cancel the dialog in the console.

The operator seems to be missing some kind of upgrade check feature that can only be triggered through the console/API.

jeff-held-aiven:
Closes #609

### DIFF
--- a/controllers/postgresql_controller.go
+++ b/controllers/postgresql_controller.go
@@ -5,20 +5,26 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/aiven/aiven-go-client/v2"
+	avngen "github.com/aiven/go-client-codegen"
+	"github.com/aiven/go-client-codegen/handler/service"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
+	pguserconfig "github.com/aiven/aiven-operator/api/v1alpha1/userconfig/service/pg"
 )
 
 // PostgreSQLReconciler reconciles a PostgreSQL object
 type PostgreSQLReconciler struct {
 	Controller
 }
+
+const waitForTaskToCompleteInterval = time.Second * 10
 
 //+kubebuilder:rbac:groups=aiven.io,resources=postgresqls,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=aiven.io,resources=postgresqls/status,verbs=get;update;patch
@@ -61,7 +67,7 @@ func (a *postgreSQLAdapter) getServiceCommonSpec() *v1alpha1.ServiceCommonSpec {
 }
 
 func (a *postgreSQLAdapter) getUserConfig() any {
-	return &a.Spec.UserConfig
+	return a.Spec.UserConfig
 }
 
 func (a *postgreSQLAdapter) newSecret(ctx context.Context, s *aiven.Service) (*corev1.Secret, error) {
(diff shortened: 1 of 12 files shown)


tests/postgresql_test.go
```diff
@@ -247,3 +249,76 @@ func TestPgCustomPrefix(t *testing.T) {
...
+	require.NoError(t, s.GetRunning(pg, pgName))
+
+	// THEN
+	// Validates instance
+	pgAvn, err := avnClient.Services.Get(ctx, testProject, pgName)
+	require.NoError(t, err)
+	// Tests non-strict yaml. By sending string-integer we expect it's parsed as a string.
+	// Default version is 15, we get 14, as we set it.
+	assert.Equal(t, "14", pgAvn.UserConfig["pg_version"])
+	assert.Equal(t, anyPointer("14"), pg.Spec.UserConfig.PgVersion)
```
byashimov, line 299: The codgen client comes with version choices https://github.com/aiven/go-client-codegen/blob/main/handler/service/service.go#L1149
Technically you can create -2 and then upgrade to -1, and that will work when 14 is EOL.

controllers/postgresql_controller.go
```diff
@@ -61,7 +67,7 @@ func (a *postgreSQLAdapter) getServiceCommonSpec() *v1alpha1.ServiceCommonSpec {
 }
 
 func (a *postgreSQLAdapter) getUserConfig() any {
-	return &a.Spec.UserConfig
+	return a.Spec.UserConfig
```
byashimov, line 70: Nice catch. Let's make this same way to other types. Is it returns a pointer to the pointer right now (in main)?

controllers/postgresql_controller.go
```diff
@@ -94,3 +100,59 @@ func (a *postgreSQLAdapter) getServiceType() string {
...
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context timeout while retrying operation, error=%q", err)
+		case <-time.After(waitForTaskToCompleteInterval):
+			val, err := f()
+			if err != nil {
+				return nil, err
+			}
+			if val != nil {
+				return val, nil
+			}
```
byashimov, line 155: It is XOR, if you get err == nil, then val is not nil.

byashimov: It shouldn't be that way, it's like a convention. If you get no error, then he value is expected. I'll leave a snippet under waitForTaskToComplete.

controllers/postgresql_controller.go
```diff
@@ -94,3 +100,59 @@ func (a *postgreSQLAdapter) getServiceType() string {
...
+	didTaskSucceed, err := waitForTaskToComplete(ctx, func() (*bool, error) {
+		t, getErr := avn.ServiceTask.Get(ctx, a.getServiceCommonSpec().Project, a.getObjectMeta().Name, task.TaskId)
+		if getErr != nil {
+			return nil, fmt.Errorf("error fetching service task %s: %q", t.Task.Id, getErr)
+		}
+
+		if t.Task.Success == nil {
+			return nil, nil
+		}
+		return t.Task.Success, nil
```
byashimov, line 130:
```suggestion
		task, err := avnGen.ServiceTaskGet(ctx, a.getServiceCommonSpec().Project, a.getObjectMeta().Name, task.TaskId)
		if err != nil {
			return nil, fmt.Errorf("error fetching pg service task %q: %w", t.Task.Id, err)
		}
		return t.Task.Success, nil
```

controllers/postgresql_controller.go
```diff
@@ -94,3 +100,59 @@ func (a *postgreSQLAdapter) getServiceType() string {
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context timeout while retrying operation, error=%q", err)
+		case <-time.After(waitForTaskToCompleteInterval):
+			val, err := f()
+			if err != nil {
+				return nil, err
+			}
+			if val != nil {
```
byashimov, line 153: OK, val can be nil (means task is still in progress). But what we really want to know if it is ready. So combining with this https://github.com/aiven/aiven-operator/pull/645/files#r1500504286

```suggestion
			if val {
				return true, nil
			}
```

controllers/postgresql_controller.go
```diff
@@ -94,3 +100,54 @@ func (a *postgreSQLAdapter) getServiceType() string {
...
+	didTaskSucceed, err := waitForTaskToComplete(ctx, func() (bool, *bool, error) {
+		t, getErr := avn.ServiceTask.Get(ctx, a.getServiceCommonSpec().Project, a.getObjectMeta().Name, task.TaskId)
+		if getErr != nil {
+			return true, nil, fmt.Errorf("error fetching service task %s: %q", t.Task.Id, getErr)
+		}
+
+		if t.Task.Success == nil {
+			return false, nil, nil
+		}
+		return true, t.Task.Success, nil
```
byashimov, line 130: Much better! Could you please return `true, *t.Task.Success, nil`.
The `nil` thing must end here, or it will make all code around check for nil to not to panic.

controllers/postgresql_controller.go
```diff
@@ -94,3 +100,59 @@ func (a *postgreSQLAdapter) getServiceType() string {
...
+	targetUserConfig := a.getUserConfig().(*pguserconfig.PgUserConfig)
+	if targetUserConfig == nil {
+		return nil
+	}
+	var targetVersion *string = targetUserConfig.PgVersion
+
+	// No need to upgrade if pg_version hasn't changed
+	if targetVersion == nil || *targetVersion == currentVersion {
+		return nil
+	}
```
byashimov, line 115: I think it better to not use a pointer string here.

```suggestion
	if targetUserConfig == nil || targetUserConfig.PgVersion == nil{
		return nil
	}

	// No need to upgrade if pg_version hasn't changed
	targetVersion := *targetUserConfig.PgVersion
	if targetVersion == currentVersion {
		return nil
	}
```

PR #647, opened 2024-02-19
feat(tests): increase timeouts for tests

jeff-held-aiven:
Individual test timeouts was increased to 20 minutes (from 10) Overall test suite timeout was increated to 90 minutes (from 42)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ test-e2e-preinstalled: check-env-vars check-avn-client ## Run end-to-end tests u
 
 test: envtest ## Run tests. To target a specific test, use 'run=TestName make test'.
 	export KUBEBUILDER_ASSETS=$(shell eval ${KUBEBUILDER_ASSETS_CMD}); \
-	go test ./tests/... -race -run=$(run) -v $(if $(run), -timeout 10m, -timeout 42m) -parallel 10 -cover -coverpkg=./controllers -covermode=atomic -coverprofile=coverage.out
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -108,7 +108,7 @@ test-e2e-preinstalled: check-env-vars check-avn-client ## Run end-to-end tests u
 
 test: envtest ## Run tests. To target a specific test, use 'run=TestName make test'.
 	export KUBEBUILDER_ASSETS=$(shell eval ${KUBEBUILDER_ASSETS_CMD}); \
-	go test ./tests/... -race -run=$(run) -v $(if $(run), -timeout 10m, -timeout 42m) -parallel 10 -cover -coverpkg=./controllers -covermode=atomic -coverprofile=coverage.out
+	go test ./tests/... -race -run=$(run) -v $(if $(run), -timeout 20m, -timeout 90m) -parallel 10 -cover -coverpkg=./controllers -covermode=atomic -coverprofile=coverage.out
```
byashimov, line 111: What 90m is for? For the whole suite? I think 60m is already too much.

Serpentiel: Good point, but why is it `42m` right now? _Probably cause it's the Answer to the Ultimate Question of Life, the Universe, and Everything..._ But shouldn't it be a different number? Perhaps, `45m` would be a better fit here.

byashimov (approved):
lgtm

PR #655, opened 2024-02-26
feat: send kube version in user-agent of aiven API clients

jeff-held-aiven:


controllers/common.go
```diff
@@ -127,12 +126,12 @@ func isAivenServerError(err error) bool {
...
 
 // NewAivenClient returns Aiven client (aiven/aiven-go-client/v2)
-func NewAivenClient(token string) (*aiven.Client, error) {
+func NewAivenClient(token, version string) (*aiven.Client, error) {
 	return aiven.NewTokenClient(token, "k8s-operator/"+version)
 }
 
 // NewAivenGeneratedClient returns Aiven generated client client (aiven/go-client-codegen)
-func NewAivenGeneratedClient(token string) (avngen.Client, error) {
+func NewAivenGeneratedClient(token, version string) (avngen.Client, error) {
```
rriski, line 134: Do we want to include the kube version or the operator version here or both? It would be good to have both versions in the user agent. Something like `k8s-operator/v0.17.0-k8s-1.26.0`. What do you think @byashimov?

byashimov: We need them both. Like in TF.

controllers/common.go
```diff
@@ -127,13 +127,13 @@ func isAivenServerError(err error) bool {
...
-func NewAivenClient(token string) (*aiven.Client, error) {
-	return aiven.NewTokenClient(token, "k8s-operator/"+version)
+func NewAivenClient(token, kubeVersion, operatorVersion string) (*aiven.Client, error) {
+	return aiven.NewTokenClient(token, fmt.Sprintf("k8s-operator/%s/%s", kubeVersion, operatorVersion))
 }
 
 // NewAivenGeneratedClient returns Aiven generated client client (aiven/go-client-codegen)
-func NewAivenGeneratedClient(token string) (avngen.Client, error) {
-	return avngen.NewClient(avngen.TokenOpt(token), avngen.UserAgentOpt("k8s-operator/"+version))
+func NewAivenGeneratedClient(token, kubeVersion string) (avngen.Client, error) {
```
rriski, line 135: The operatorVersion seems to be missing here, error in basic_controller.go:

image

rriski (approved):
LGTM, although now we have two places where `operatorVersion` is specified. It would be good to have a single source of truth.

https://github.com/aiven/aiven-operator/pull/655/files#diff-96e66e1ccb0083ac5a2b7fc37cb002803d242ea67c1981a5d09ef8ee2668354cR38-R40 + https://github.com/aiven/aiven-operator/pull/655/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R32-R34

Serpentiel (changes requested):
As @rriski mentioned, we need to get rid of the additional definition of the provider version.

We also should make the approach similar to the one in our Terraform Provider, here's how it works:

- `version` is defined somewhere, [preferably in `main.go`](https://github.com/aiven/terraform-provider-aiven/blob/main/main.go#L21)
- It is then overwritten during the compile time via `go build`'s `ldflags` parameter, similarly to [this approach](https://github.com/aiven/terraform-provider-aiven/blob/main/.goreleaser.yml#L8).

TL;DR. While I think it's unreasonable to add GoReleaser support in this PR, I think it'll be beneficial for us to define the `operatorVersion` in one place only; this could be done by setting the one in `suite_test.go` to always be a constant string `test`, like we do [over here](https://github.com/aiven/terraform-provider-aiven/blob/main/internal/sdkprovider/provider/provider_test.go#L8). This also allows to understand, via the `User-Agent`, which requests originated from our tests.

Serpentiel (approved):
lgtm

PR #676, opened 2024-03-18
refactor: use compatible client error checks

byashimov:
Now we are mixing the old and the new client. Error checks must support both types of errors: the old and the new one.


byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/8324234778)

PR #685, opened 2024-03-27
feat(clickhousedatabase): remove termination protection

byashimov:
The termination protection is not an API feature.
Since it might be misleading, it is better to remove it.
Which is also allows us to remove the webhook.


Serpentiel:
@byashimov we have similar pseudo "termination protection" functionality in TF, do you think it worths removing it as well?

byashimov:
> @byashimov we have similar pseudo "termination protection" functionality in TF, do you think it worths removing it as well?

This resource is not released yet. So it won't break things.
But removing it from TF will be a breaking change.

Serpentiel (approved):
lgtm

PR #696, opened 2024-04-02
chore: fix release flow

Serpentiel:


main.go
```diff
@@ -30,7 +30,7 @@ var (
 )
 
 // operatorVersion is the current version of the operator.
-const operatorVersion = "0.17.0"
+const operatorVersion = "dev"
```
Serpentiel, line 33: This is similar to the approach we do in Terraform

byashimov, line 33: I think a constant can't change, please have tested it.

byashimov (approved):
lgtm

PR #707, opened 2024-04-08
refactor: share project and service fields

byashimov:
- Fixes project regex pattern
- ServceName and Project fields are shared across all types
- Both fields are marked as immutable


byashimov:
Reviewers, please validate regexps.

charts/aiven-operator-crds/templates/aiven.io_cassandras.yaml
```diff
@@ -139,8 +139,8 @@ spec:
                   maxLength: 128
                   type: string
                 project:
-                  description: Target project.
-                  format: ^[a-zA-Z0-9_-]*$
+                  description: Identifies the project this resource belongs to
+                  format: ^[a-zA-Z0-9_-]+$
```
Serpentiel, line 143: This can't be an empty string, right?

byashimov: Can't imagine why project name would be empty. The field is required, though.

Serpentiel (approved):
lgtm

PR #719, opened 2024-04-30
chore(ci): update commitlint config to v19 format and update commitlint-github-action

rriski:
We run commitlint locally via trunk, but trunk-action doesn't run commitlint in CI so we still need `wagoid/commitlint-github-action`.


byashimov (approved):
lgtm

PR #728, opened 2024-05-07
fix(redis): remove REDIS_CA_CERT secret key

byashimov:
The cert can't be used with this type of service.


byashimov:
The cert is also hidden in the Console.

byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/8983208325)

PR #734, opened 2024-05-14
fix(connectionpool): check user exists

byashimov:
- Adds service user existence check
- Moves test yaml to the examples


controllers/connectionpool_controller.go
```diff
@@ -219,24 +215,25 @@ func (h ConnectionPoolHandler) checkPreconditions(ctx context.Context, avn *aive
...
-			if isNotFound(err) {
-				return false, nil
-			}
-			return false, err
-		}
+	if !isRunning {
+		return false, nil
+	}
+
+	_, err = avn.Databases.Get(ctx, cp.Spec.Project, cp.Spec.ServiceName, cp.Spec.DatabaseName)
```
byashimov, line 227: There is no single user get method. The GET endpoint returns a list. Using the original client.

byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/9075923205)

controllers/connectionpool_controller.go
Serpentiel: Is this a potential bug to work on in the codegen client?

byashimov: It's not a bug, there is just no such an endpoint in the API.

Serpentiel: I think we need to raise this to the relevant stakeholders, because this isn't a first time when this happens. I would recommend making up a list of all of these instances where this is needed.

rriski (approved):
LGTM

PR #768, opened 2024-06-25
docs: generate usage commands

byashimov:
Generates example usage commands from examples.
Forms `kubectl get` output with true values.


byashimov:
image

rriski (approved):

PR #778, opened 2024-07-02
fix(deps): go client string pointers update

byashimov:
Updated client brings pointers to strings.


byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/9759108875)

rriski (approved):

PR #786, opened 2024-07-10
fix(tests): assert destroy errors

byashimov:


byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/9876140700)

PR #787, opened 2024-07-12
fix(clickhousegrant): revoke known privileges only

byashimov:
- Allows to create grants for non-existing table (Clickhouse can do that)
- Saves the state to revoke only granted privileges.
  Resolves concurrency access to the grants table in Clickhouse.


byashimov:
Tests https://github.com/aiven/aiven-operator/actions/runs/9940577849

PR #796, opened 2024-07-24
fix(serviceintegration): sends empty strings

byashimov:
The string pointers in the [updated client](https://github.com/aiven/aiven-operator/pull/778) brake service integrations. The latest update on BE doesn't ignore empty values.


byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/10073408312)

PR #797, opened 2024-07-24
fix(kafkaschema): poll 404

### LINKED ISSUE #800: feat(kafkaschema): add schemaType to allow more then AVRO
hi

atm only AVRO schemas can be managed via the operator. 

Would be great if we could also manage JSON and protobuf schemas, maybe via a `schemaType` (suggestion, same as go api client)

byashimov:
- Fix `KafkaSchema`: poll resource availability
- Add `KafkaSchema` field `schemaType`, type `string`: Schema type

Resolves #800


byashimov:
[tests](https://github.com/aiven/aiven-operator/actions/runs/10112393633)

byashimov:
Fixed conflicts.

PR #813, opened 2024-09-05
fix: pg upgrade check test

rriski:
Don't test against the latest version, it may not be supported yet by the operator


rriski:
[tests](https://github.com/aiven/aiven-operator/actions/runs/10715609288)

byashimov (approved):

PR #836, opened 2024-10-29
feat: add autoscaler service integration support

rriski:
* Adds `autoscaler` service integration and service integration endpoint support


rriski:
[tests](https://github.com/aiven/aiven-operator/actions/runs/11591400691)

byashimov:
Hey. Looks good to me, few things I would do:

1. put some validation into service update part, here [generic_service_handler.go](https://github.com/aiven/aiven-operator/blob/de975aa80dd0a2176b6012b806c2d2fd395a1cb3/controllers/generic_service_handler.go#L93-L93):
```go
diskSpace := v1alpha1.ConvertDiskSpace(o.getDiskSpace())
if diskSpace > 0 {
	for _, v := range oldService.ServiceIntegrations {
		if v.IntegrationType == service.IntegrationTypeAutoscaler {
			return fmt.Errorf("cannot set disk space for service with autoscaler integration enabled")
		}
	}
}
```
2. Add a warning here on the disk space field description [common.go](https://github.com/aiven/aiven-operator/blob/de975aa80dd0a2176b6012b806c2d2fd395a1cb3/api/v1alpha1/common.go#L166-L166), something like "should not be set while the autoscaler service integration is running"

Additionally we could have webhook validation, but the setup manager does not passthrough the go client. So let's not do that.

byashimov (changes requested):
Please add disk_space vs autoscaler validation 🙏

PR #848, opened 2024-11-19
feat: add Flink kind

rriski:
Adds new kind `Flink`. Most of the changeset is generated with:

```bash
operator-sdk create api --version v1alpha1 --kind Flink --force
operator-sdk create webhook --version v1alpha1 --kind Flink --defaulting --programmatic-validation --conversion
```


rriski:
[tests](https://github.com/aiven/aiven-operator/actions/runs/11909135431)

byashimov (changes requested):
Hey. Could you please add an "example", like [this](https://github.com/aiven/aiven-operator/blob/main/docs/docs/api-reference/examples/cassandra.yaml) one?
It will render then a ["usage"](https://aiven.github.io/aiven-operator/api-reference/cassandra.html#usage-example) block with a beginner callout (create token, etc), and useful commands/examples.
Then someday we can get rid of [this page](https://aiven.github.io/aiven-operator/resources/project.html).
The example can be used as a test too, but I leave this up to you.

byashimov (approved):

PR #879, opened 2025-01-31
refactor(redis): add Redis to Valkey instructions link

byashimov:
- Adds the deprecation to the changelog
- Adds upgrade instructions link

PR #882, opened 2025-02-03
docs(redis): add EOL notice to Redis resource [DOC-1302]

staceysalamon-aiven:
Adds EOL notice to Redis resource and updates language in API reference notice. Also replaces note callouts
across all docs with a prerequisites section
to make notices like this EOL callout more prominent.

### DIFF
--- a/docs/docs/resources/redis.md
+++ b/docs/docs/resources/redis.md
@@ -6,9 +6,14 @@ weight: 50
 
 Aiven for Redis®\* is a fully managed in-memory NoSQL database that you can deploy in the cloud of your choice to store and access data quickly and efficiently.
 
-!!! note
-    Before going through this guide, make sure you have a [Kubernetes cluster](../installation/prerequisites.md) with the operator installed (see instructions for [helm](../installation/helm.md) or [kubectl](../installation/kubectl.md)),
-    and a [Kubernetes Secret with an Aiven authentication token](../authentication.md).
+!!! warning "End of life notice"
+
+    The Aiven for Caching offering (formerly Aiven for Redis®) is entering its [end-of-life cycle](https://aiven.io/docs/platform/reference/end-of-life). From **February 15th, 2025**, it will not be possible to start a new Aiven for Caching service, but existing services up until version 7.2 will still be available until end of life. From **March 31st, 2025**, Aiven for Caching will no longer be available and all existing services will be migrated to Aiven for Valkey™. You can [upgrade to Valkey for free yourself](https://aiven.io/docs/products/caching/howto/upgrade-aiven-for-caching-to-valkey) before then.
+
+## Prerequisites
+
+* A Kubernetes cluster with Aiven Kubernetes Operator installed using [helm](../installation/helm.md) or [kubectl](../installation/kubectl.md).
+* A [Kubernetes Secret with an Aiven authentication token](../authentication.md).
 
 ## Create a Redis instance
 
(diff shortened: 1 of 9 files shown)


docs/docs/resources/redis.md
```diff
@@ -6,9 +6,14 @@ weight: 50
...
 
-!!! note
-    Before going through this guide, make sure you have a [Kubernetes cluster](../installation/prerequisites.md) with the operator installed (see instructions for [helm](../installation/helm.md) or [kubectl](../installation/kubectl.md)),
-    and a [Kubernetes Secret with an Aiven authentication token](../authentication.md).
+!!! warning "End of life notice"
+
+    In March 2024, a new licensing model was announced for Redis® that impacts the Aiven for Caching offering (formerly Aiven for Redis®). Aiven for Caching is entering its end-of-life cycle to comply with Redis's copyright and license agreements. From **February 15th, 2025**, it will not be possible to start a new Aiven for Caching service, but existing services up until version 7.2 will still be available until end of life. From **March 31st, 2025**, Aiven for Caching will no longer be available and all existing services will be migrated to Aiven for Valkey™. You can [upgrade to Valkey for free yourself](https://aiven.io/docs/products/caching/howto/upgrade-aiven-for-caching-to-valkey) before then.
+
```
rriski, line 12: Hey Stacey, we added a deprecation warning here: https://github.com/aiven/aiven-operator/pull/879

It shows up in docs like this:

image


I think we could remove this part.

rriski (changes requested):

docs/docs/resources/redis.md
staceysalamon-aiven: Hey, yeah, I saw that, but I'd like to include this on both Redis pages and at least have the key dates here. I also wanted to improve the link to the upgrade instructions. I can remove the first sentence though.

docs/docs/resources/service-integrations.md
```diff
@@ -10,9 +10,10 @@ See
 our [Getting Started with Service Integrations guide](https://aiven.io/docs/platform/concepts/service-integration)
 for more information.
 
-!!! note
-    Before going through this guide, make sure you have a [Kubernetes cluster](../installation/prerequisites.md) with the operator installed (see instructions for [helm](../installation/helm.md) or [kubectl](../installation/kubectl.md)),
-    and a [Kubernetes Secret with an Aiven authentication token](../authentication.md).
+## Prerequisites
+
+* A Kubernetes cluster with Aiven Kubernetes Operator [installed](../installation/helm.md).
+* A [Kubernetes Secret with an Aiven authentication token](../authentication.md).
```
rriski, line 16: I think the before looks better, wdyt @staceysalamon-aiven? Additionally mysql and redis docs still use the note mkdoc [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/). The new copy doesn't link to installation/kubectl.md (except in cassandra and mysql docs which aren't updated. this change doesn't seem to be in scope of the PR description?

## Before

image

## After

image

## Proposal

Is removing the link to kubectl docs and admonition required? Why not keep the admonition and update copy?

image

rriski (changes requested):
Mysql and cassandra resource docs use the admonition while others are updated. See my proposal in the comment.

docs/docs/resources/service-integrations.md
staceysalamon-aiven: You're right, I missed MySQL - thanks! I've updated that one too from the quote-style text to a prerequisites section. All of these are now lists of prerequisites to align with our docs standards and also to make the page less noisy, making important callouts like the EOL notice more prominent. 

The Cassandra doc is being updated in #885 since a notice also needs to be added to that page, but published a little later.

I've also added back the links to the separate instructions for helm and kubectl.

rriski (approved):

PR #891, opened 2025-02-12
docs(valkey): Add Valkey resource example

bogdanp05:
I added a resource example for Valkey.
I considered removing the example for Redis, but I left it in because it has the EOL warning. I think we can safely remove it after the EOL date, but let me know if we should remove it already.


docs/docs/resources/valkey.md
```diff
@@ -0,0 +1,131 @@
...
+  # gets the authentication token from the `aiven-token` Secret
+  authSecretRef:
+    name: aiven-token
+    key: token
+
+  # outputs the Valkey connection on the `valkey-secret` Secret
+  connInfoSecretTarget:
+    name: valkey-secret
+
+  # add your Project name here
```
staceysalamon-aiven, line 33: In general, we avoid adding instructions within the code comments. I'd add the placeholder after the code block similarly to how it's done for Clickhouse:  https://github.com/aiven/aiven-operator/blob/a1cda477b850ed4d574422c688aa3dff1ad53949/docs/docs/resources/clickhouse.md?plain=1#L40

docs/docs/resources/valkey.md
```diff
@@ -0,0 +1,131 @@
...
+Data
+====
+SSL:       8 bytes
+USER:      7 bytes
+HOST:      60 bytes
+PASSWORD:  24 bytes
+PORT:      5 bytes
+```
+
+You can use the [jq](https://github.com/jqlang/jq) to quickly decode the Secret:
```
staceysalamon-aiven, line 115:
```suggestion
You can use [jq](https://github.com/jqlang/jq) to quickly decode the Secret:
```

staceysalamon-aiven (approved):
Thanks for adding this example! I made a couple of small suggestions to the text. I also agree that we should wait for EOL to remove Redis.

staceysalamon-aiven (approved):
LGTM

byashimov (approved):

PR #894, opened 2025-02-17
feat: add `databaseName` to `Database` kind

### LINKED ISSUE #874: Database name customization
Hi there,

It is not possible to create a PostgreSQL database which complies to the [PostgreSQL conventions](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) (simply put, it includes an **underscore**) because the Aiven's `Database` kubernetes object is using its own name (which must be RFC1123 compliant) as name for the actual database being created inside the Aiven service.

I would like to have the possibility of overwriting the actual name of the database with a `name` property or similar.

Thanks in advance

rriski:
Fixes #874. 

PostgreSQL identifiers can have underscores and even emojis. However,
Aiven's API allows up to 40 character long names but doesn't accept emojis.

K8s metadata name uses RFC1123 compliant names which is stricter. For example
underscores (_) aren't supported in metadata names.

Add `databaseName` field to `Database` kind's spec which is preferred over metadata
name if non RFC 1123 compliant database names are desired.

See `ClickhouseDatabase` for a similar implementation.


rriski:
[tests](https://github.com/aiven/aiven-operator/actions/runs/13367813438)

api/v1alpha1/database_types.go
```diff
@@ -26,6 +27,12 @@ type DatabaseSpec struct {
 	// from being deleted by Kubernetes. It is recommended to enable this for any production
 	// databases containing critical data.
 	TerminationProtection *bool `json:"terminationProtection,omitempty"`
+
+	// +kubebuilder:validation:MaxLength=40
```
rriski, line 31: 🔎 Postgres default is 63 AFAIK but we enforce a stricter limit on the backend

api/v1alpha1/database_types.go
```diff
@@ -26,6 +27,12 @@ type DatabaseSpec struct {
...
+	// +kubebuilder:validation:Pattern="^[a-zA-Z0-9_][a-zA-Z0-9_-]{0,39}$"
```
rriski, line 32: 🔎 Again, this is a stricter limit on the backend than the PG default

api/v1alpha1/database_types.go
```diff
@@ -26,6 +27,12 @@ type DatabaseSpec struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
rriski, line 33: 🔎 `databaseName` is immutable, to create a new database use a new resource

api/v1alpha1/database_types.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 // DatabaseSpec defines the desired state of Database
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.databaseName) == has(self.databaseName)",message="databaseName can only be set during resource creation."
```
rriski, line 10: 🔎 Disallow setting or removing `databaseName` after resource creation

byashimov (approved):

byashimov (approved):

PR #925, opened 2025-05-15
chore(go): upgraded golangci-lint version

vmyroslav:
- upgraded golangci-lint version


tests/suite_test.go
```diff
@@ -64,8 +64,10 @@ func TestMain(m *testing.M) {
 		log.Fatal(err)
 	}
 
-	defer teardownSuite(env)
-	os.Exit(m.Run())
+	exitCode := m.Run()
+	teardownSuite(env)
+
+	os.Exit(exitCode)
```
byashimov, line 70: I think the behaviour has changed: defer used to teardown the env even after a panic, but not now.

utils/alloydbomni/service_account_credentials_validator.go
```diff
@@ -1,4 +1,4 @@
-package alloydbomniUtils
+package alloydbomniutils
```
byashimov, line 1: Package name should be in lowercase.

utils/clickhouse/clickhouse_grant.go
```diff
@@ -1,4 +1,4 @@
-package chUtils
+package chutils
```
byashimov, line 1: Package name should be in lowercase.

tests/suite_lib_test.go
```diff
@@ -6,12 +6,14 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/stretchr/testify/assert"
 )
```
byashimov, line 12: Please run `make fmt`.

byashimov (changes requested):

utils/alloydbomni/service_account_credentials_validator.go
vmyroslav: But it is in lowercase

tests/suite_test.go
vmyroslav: Yes, but before `teardownSuite` was never executed as the process was terminated with `os.Exit`

tests/suite_lib_test.go
vmyroslav: done

tests/suite_test.go
vmyroslav: upd

byashimov (approved):

PR #936, opened 2025-06-02
refactor(database): use gen client

byashimov:
Replaces the old client with generated.

Resolves NEX-1560.


byashimov:
```
--- PASS: TestDatabase_databaseName (220.48s)
```

PR #945, opened 2025-06-13
feat: added support for powering services on and off

byashimov:
Resolves NEX-1149.


byashimov:
tests https://github.com/aiven/aiven-operator/actions/runs/15776915050

byashimov:
> I tested it a few services. LGTM. The only thing that come to my mind as an improvement is to add a proper condition and remove existing conditions from depending services. Something like this:
> 
> ```go
> 	isOperational, err := checkServiceIsOperational(ctx, avnGen, g.Spec.Project, g.Spec.ServiceName)
> 	if err != nil {
> 		if errors.Is(err, errServicePoweredOff) {
> 			meta.SetStatusCondition(&g.Status.Conditions, metav1.Condition{
> 				Type:    "ServicePoweredOff",
> 				Status:  metav1.ConditionTrue,
> 				Reason:  "ParentServicePoweredOff",
> 				Message: fmt.Sprintf("ClickHouse service %s/%s is powered off. Power on the service to enable grant management.", g.Spec.Project, g.Spec.ServiceName),
> 			})
> 			meta.RemoveStatusCondition(&g.Status.Conditions, conditionTypeInitialized)
> 			meta.RemoveStatusCondition(&g.Status.Conditions, conditionTypeRunning)
> 			meta.RemoveStatusCondition(&g.Status.Conditions, ConditionTypeError)
> 			return false, nil
> 		}
> 		return false, err
> 	}
> 	if !isOperational {
> 		return false, nil
> 	}
> ```

Hey. The proposed solution completely breaks dependant resource that is updated (that's what the doc says too). I mean, suppressing `errServicePoweredOff` and adding a new condition is more complicated solution, because it involves all the kinds we have excluding the services. I'm proposing to keep this simple till we learn we need this functionality.

PR #950, opened 2025-06-18
fix(kafkaconnector): better handle api errors

byashimov:
Resolves NEX-1610.

- Doesn't call `ServiceKafkaConnectGetConnectorStatus` to know if the connector exists. 
  It might return OK, and `ServiceKafkaConnectCreateConnector` might fail with 404 right after. Instead calls "create", and `ServiceKafkaConnectEditConnector` to update it if gets a conflict.
- Additionally, retries all 5xx errors.


byashimov:
tests https://github.com/aiven/aiven-operator/actions/runs/15732063195

vmyroslav (approved):

PR #978, opened 2025-07-25
feat(kafkatopic): added concurrent reconcilers

### LINKED ISSUE #974: Mass topic creation is very slow (~1 topic per minute)
I have created a Helm Chart to manage the following Aiven Kafka custom resources.

- Users (ServiceUser)
- Topics (KafkaTopic)
- ACLs (KafkaACL)

We are trying to create approx. 20 users, 250 Topics and 300 ACLs with this Helm chart.

The creation of users and ACLs went in an acceptable time frame (30 seconds for all users, 3 minutes for all ACLs), but topic creation seemed to progress very slowly. The operator created only one topic at a time, and required 60-62 seconds for each. For 250 topics, **this took over 4 hours** to complete. 

What I did to measure this: 

```
kubectl get kafkatopic -o custom-columns=NAME:.spec.topicName,STATUS:.status.state --watch | ts
```

Every minute, this emitted two lines per topic, like 

```
Jul 18 17:38:25 Test.Messages.UnblockConsumer                                                           <none>
Jul 18 17:38:25 Test.Messages.UnblockConsumer                                                           ACTIVE
```

The deletion of the same amount of items went much quicker again, within a few minutes everything was gone.

While we will not create lots of topics regularly, 250 is our initial set of topics for a new cluster, and we would prefer this to be much quicker, so we can deploy the Helm chart with flux, without having to define a timeout of hours.

The delay of almost exactly one minute (60-62 seconds) seems like a deliberate delay. Is there anything to speed up this process?

Best regards, Markus

vmyroslav:
### Optimize KafkaTopic reconciliation performance
Resolves #974  

### Results:
  - ~60 seconds for 250 topics
  - singleflight deduplicate ~80% of the call, which still would be close to O(N), but significantly reduced globally with number of retries

### DIFF
--- a/controllers/kafkatopic_controller.go
+++ b/controllers/kafkatopic_controller.go
@@ -10,11 +10,13 @@ import (
 	avngen "github.com/aiven/go-client-codegen"
 	"github.com/aiven/go-client-codegen/handler/kafkatopic"
 	"github.com/aiven/go-client-codegen/handler/service"
+	"golang.org/x/sync/singleflight"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


byashimov (approved):
Nice job!

Issue #970, opened 2025-07-18
Ability to disable unused controllers


mikkoc:
Hello,
I would like to request two nice-to-have features.

1. Allow to either disable certain controllers, or only enable certain controllers (preferred). We only use Kafka, so we're not interested in all the other stuff that the Operator supports. Expose this ability via Helm.
2. Allow to configure LOG_LEVEL of the Operator via Helm values.

Thanks

mikkoc:
I realise that the 1st request might be partially covered by https://github.com/aiven/aiven-operator/issues/380

PR #982, opened 2025-07-31
feat: added customizable log levels

vmyroslav:

Resolves: NEX-1700.
Contributes to #970 

- added customizable log levels for the customer

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 
+	"go.uber.org/zap/zapcore"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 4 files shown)


vmyroslav:
[Tests](https://github.com/aiven/aiven-operator/actions/runs/16646549642)

main.go
```diff
@@ -57,14 +59,27 @@ func main() {
...
 	flag.BoolVar(&development, "development", true, "Configures the logger to use a development config (stacktraces on warnings, no sampling)")
+
 	opts := zap.Options{
 		Development: development,
 	}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	// set log level from environment variable if provided and no flag was set
+	if logLevel := os.Getenv("LOG_LEVEL"); logLevel != "" && opts.Level == nil {
```
byashimov, line 70: This line sends a flag https://github.com/aiven/aiven-operator/pull/982/files#diff-493d41a6cf8a4d52b86d40c55fb49d39b69c3f29a9042406b17f8a0a2056128aR67
We shouldn't expect an env variable here.

main.go
```diff
@@ -148,3 +163,19 @@ func main() {
...
+	case "info":
+		return zapcore.InfoLevel, true
+	case "warn", "warning":
+		return zapcore.WarnLevel, true
+	case "error":
+		return zapcore.ErrorLevel, true
+	default:
+		return zapcore.InfoLevel, false // default to info if invalid level provided
+	}
+}
```
byashimov, line 181: I think `if err != nil` is a common practice. What is the reason of returning a bool instead of an error directly?

main.go
```diff
@@ -57,14 +59,27 @@ func main() {
...
+		if !valid {
+			_, _ = fmt.Fprintf(os.Stderr, "Warning: Invalid LOG_LEVEL value '%s', using default level. Valid values: debug, info, warn, warning, error\n", logLevel)
+		} else {
+			opts.Level = level
+		}
+	}
+
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	ctrl.Log.Error(errors.New("test error"), "This is a test error message to verify logging setup")
```
byashimov, line 81: So, if it works it will output an error telling it is not an error. But if it is not?
1. Should we include an instruction that tells user must expect this output to prove the operator was setup correctly?
2. Can we use debug or info level here instead?
3. Could it be the final step of the function and output something like "all good"?

byashimov (changes requested):

main.go
byashimov: I mean, why then we need `zap-log-level`? It is not handled. Am I missing something? :(

main.go
vmyroslav: Discussed

main.go
vmyroslav: discussed

PR #986, opened 2025-08-04
feat(secret): added secret watcher

vmyroslav:

Resolves: NEX-1703.
- added controller to watch changes in `connInfoSource` and trigger reconciliation in dependent resources


tests/secret_watcher_test.go
```diff
@@ -0,0 +1,474 @@
...
+  authSecretRef:
+    name: aiven-token
+    key: token
+
+  project: %[1]s
+  cloudName: %[6]s
+  plan: startup-16
+
+  userConfig:
+    pg_version: "15"
```
byashimov, line 393: Please remove the version: as far as I can tell it is not important for the test. Someday this one will reach the EOL and the test will fail.

controllers/secret_watch_controller.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := c.Update(ctx, latest); err != nil {
+		return fmt.Errorf("failed to update resource annotation: %w", err)
+	}
+
+	c.Log.Info("triggered reconciliation for resource",
+		"resource", types.NamespacedName{Name: latest.GetName(), Namespace: latest.GetNamespace()},
+		"kind", latest.GetObjectKind().GroupVersionKind().Kind)
+
+	return nil
+}
```
byashimov, line 240: Please put the annotations into constants. They are shared with the testing code at least.

controllers/secret_watch_controller.go
```diff
@@ -0,0 +1,240 @@
...
+	annotations["controllers.aiven.io/secret-source-updated"] = fmt.Sprintf("%d", time.Now().Unix())
+
+	// clear the processed generation annotation to force the basic controller to reconcile
+	delete(annotations, "controllers.aiven.io/generation-was-processed")
+
+	latest.SetAnnotations(annotations)
+
+	if err := c.Update(ctx, latest); err != nil {
+		return fmt.Errorf("failed to update resource annotation: %w", err)
+	}
```
byashimov, line 233: What happens if user applies two yamls simultaneously: user and secret (assuming both exist)? Would this update fight with the user controller?

tests/secret_watcher_test.go
```diff
@@ -0,0 +1,474 @@
...
+  authSecretRef:
+    name: aiven-token
+    key: token
+
+  project: %[1]s
+  cloudName: %[5]s
+  plan: startup-16
+
+  userConfig:
+    pg_version: "15"
```
byashimov, line 135: Please remove the version if it is not important.

controllers/secret_watch_controller.go
vmyroslav: done

tests/secret_watcher_test.go
vmyroslav: removed

controllers/secret_watch_controller.go
vmyroslav: What can happen.
  - secret watcher and main controller update the same resource simultaneously
  - both try to modify annotations (secret watcher: clears processedGenerationAnnotation, main controller: sets it)
  - K8S concurrency control detects resource version mismatch

It is not critical for the result, however I added additional check `errors.IsConflict(err)` to avoid cluttering logs with misleading errors. I also added few tests to illustrate this situation. Thanks for spotting that.

PR #987, opened 2025-08-06
refactor: consolidate status conditions

byashimov:
Resolves NEX-1670.

Consolidates status conditions `Create` and `Update` into `CreateOrUpdate` due to limitations in reliably determining operation type.

1. The go client's retry mechanism may create an instance successfully but receive a 5xx error,
causing the next attempt to get a conflict error
2. Remote state may be temporarily inconsistent, so GET checks can return false positives/negatives
3. Some handlers implement their own retry logic and keep trying to create instances until success

Therefore, we can't rely on "exists" checks to determine the operation type.

### DIFF
--- a/controllers/basic_controller.go
+++ b/controllers/basic_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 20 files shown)


controllers/basic_controller.go
```diff
@@ -486,6 +488,35 @@ func (i *instanceReconcilerHelper) createOrUpdateInstance(ctx context.Context, o
 		"annotations", o.GetAnnotations(),
 	)
 
+	// We don't really know if an instance is created or updated because:
+	// 1. The go client's retry mechanism may create an instance successfully but receive a 5xx error,
+	//    causing the next attempt to get a conflict error
+	// 2. Remote state may be temporarily inconsistent, so GET checks can return false positives/negatives
+	// 3. Some handlers implement their own retry logic and keep trying to create instances until success
+	// Therefore, we can't rely on "exists" checks to determine the operation type
+	const reason = "CreateOrUpdate"
```
vmyroslav, line 497: nit: Conditions looks to me like events, meaning that maybe it makes sense to name then in past tense as well

vmyroslav (dismissed):

vmyroslav (approved):

PR #991, opened 2025-08-11
docs(aiven_token): added docs for aiven token configuration

vmyroslav:

Contributes: NEX-1641
- added some documentation for token management in operator, later we will add watcher for token updates


vmyroslav:
I updated the `RBAC` section and removed the comments from bash code examples.

byashimov (approved):

PR #992, opened 2025-08-13
fix: fixed ConnectionPool tests to use proper secret keys

vmyroslav:

Fixes: NEX-747
- use proper secret keys in ConnectionPool tests


vmyroslav:
[Tests](https://github.com/aiven/aiven-operator/actions/runs/16934409814)

byashimov (approved):

PR #995, opened 2025-08-15
feat: use shared resources to run tests

byashimov:
Resolves NEX-1743.

Shared test resources can save costs and time to run tests.

- Adds shared test resources storage
- Uses a shared PG and Clickhouse to run user/db tests
- Groups tests by tags


byashimov:
Got import issues in ACC tests, moving to draft.

PR #1000, opened 2025-08-20
fix: added retry on conflict to the test session

vmyroslav:

- added retry to the test session to avoid conflicts in case of multiple resources modification. Outside of the tests it should not be an issue, as kubectl has a [retry](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/patcher.go#L364) logic as well on its side


vmyroslav:
[Tests](https://github.com/aiven/aiven-operator/actions/runs/17092767113)

PR #1001, opened 2025-08-20
feat: added addon install on GKE cluster

vmyroslav:

Resolves: NEX-1747
- added Aiven operator installation to GKE with Taskfile commands for orchestration

The REAME guide and other sugar will be added in the next PR


vmyroslav:
Makes sense. I'll do both changes during the next PR related to this topic

PR #1006, opened 2025-08-26
fix(tests): resolve parallel test resource conflicts in tests

vmyroslav:

Resolves: NEX-1782

**Problem**:
  - `TestClickhouseGrant` and `TestClickhouseGrantExample` both used hardcoded resource name "`writer`" for `ClickhouseRole`
  - When running in parallel (t.Parallel()), both tests created resources with identical names in the same Kubernetes cluster
  - This caused race conditions where one test would overwrite another's resources
 
Additionaly:
  - In secret watcher patch implementation using more idiomatic `client.MergeFrom()` 
  - Fixed conflicts with shared resources in tests

### DIFF
--- a/controllers/secret_watch_controller.go
+++ b/controllers/secret_watch_controller.go
@@ -4,7 +4,6 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"reflect"
 	"time"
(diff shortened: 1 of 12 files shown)


vmyroslav:
Tests: [1](https://github.com/aiven/aiven-operator/actions/runs/17247125540), [2](https://github.com/aiven/aiven-operator/actions/runs/17249881561)

PR #1148, opened 2026-04-20
feat(mysql/postgres): added migrationSecretSource

vmyroslav:

- Added `migrationSecretSource` field to PostgreSQL and MySQL, allowing migration credentials to be stored in a Kubernetes Secret instead of plain text in the CR spec
- Secret keys must match `userConfig.migration` JSON field names (e.g., host, port, password, ssl)
- When `migrationSecretSource` is set, it fully replaces userConfig.migration — no merging
- Webhook warnings are emitted when both migrationSecretSource and userConfig.migration are provided

The PR is split for 2 parts. Next part should contain:
- `deleteAfterMigration` option for automatic Secret cleanup after migration completes
- migration status polling for inline` userConfig.migration` (currently only works with migrationSecretSource)

Resolves: NEX-2432


controllers/generic_service_handler.go
```diff
@@ -260,6 +270,53 @@ func (h *genericServiceHandler) get(ctx context.Context, avnGen avngen.Client, o
...
+				Type:               v1alpha1.ConditionTypeMigrationComplete,
+				Status:             metav1.ConditionTrue,
+				ObservedGeneration: o.getObjectMeta().Generation,
+				Reason:             v1alpha1.MigrationReasonDone,
+				Message:            "Migration completed (no active migration found)",
+			})
+			return
+		}
+
+		return
```
byashimov: The error is silently swallowed (when not 404). Is that intended?
Please leave a comment if so.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
+- Add `MySQL` field `migrationSecretSource`, type `object`: Reference to a Secret containing migration
+  credentials. Secret keys must match userConfig.migration JSON field names. If set, takes precedence over userConfig.migration.
+
```
byashimov, line 7: Add PG please.

controllers/mysql_controller.go
```diff
@@ -85,6 +91,78 @@ func (a *mySQLAdapter) newSecret(_ context.Context, s *service.ServiceGetOut) (*
...
+		if err != nil {
+			return fmt.Errorf("migration secret %q: invalid reestablish_replication value %q: %w", ref.Name, v, err)
+		}
+		m.ReestablishReplication = lo.ToPtr(b)
+	}
+
+	if a.Spec.UserConfig == nil {
+		a.Spec.UserConfig = &mysqluserconfig.MysqlUserConfig{}
+	}
+	a.Spec.UserConfig.Migration = m
```
byashimov: Won't it happen, that the secret will be available in "kubectl describe"? 
This looks bothering.

controllers/generic_service_handler.go
vmyroslav: Updated. Return an error.

CHANGELOG.md
vmyroslav: Done

controllers/mysql_controller.go
vmyroslav: Thanks for that. Changed the interface for that to avoid storing it in a spec.

controllers/migration_secret.go
```diff
@@ -0,0 +1,28 @@
...
+// and returns its data as trimmed strings keyed by secret key name.
+func readMigrationSecret(ctx context.Context, k8s client.Reader, namespace string, ref *v1alpha1.MigrationSecretSource) (map[string]string, error) {
+	secret := &corev1.Secret{}
+	if err := k8s.Get(ctx, types.NamespacedName{Name: ref.Name, Namespace: namespace}, secret); err != nil {
+		return nil, fmt.Errorf("cannot get migration secret %q: %w", ref.Name, err)
+	}
+
+	data := make(map[string]string, len(secret.Data))
+	for k, v := range secret.Data {
+		data[k] = strings.TrimSpace(string(v))
```
byashimov, line 25: nit: This looks valid, and I like it. Just fyi: PG supports leading space in passwords. So I would just mention in the doc string that we trim spaces from values.

controllers/generic_service_handler.go
```diff
@@ -32,6 +32,15 @@ func (h *genericServiceHandler) createOrUpdate(ctx context.Context, avnGen avnge
 		return err
 	}
 
+	userCfg := o.getUserConfig()
+	if mp, ok := o.(migrationSecretProvider); ok {
```
byashimov, line 36:
```suggestion
	if mp, ok := o.(migrationSecretProvider); ok && mp.getMigrationSecretSource() != nil {
```
The line 243 makes this better, copied from there.

byashimov (approved):
Overall looks good to me, though:

1\. I would make `getUserConfigWithMigration` accept the secret to merge. 
Then the `readMigrationSecret` would be outside since it is a shared logic and doesn't have to be in the adapter itself.

2\. I think the marshaling one-by-one field is tricky to maintain, and easy to fail with hard-coded field names. We used to have this issue with the old client — each time you go to the API docs and compare symbol by symbol, that just doesn't work.

I would probably go for struct inspection and dynamic type conversion, custom `UnmarshalJSON` or you can try to do this:

```go
Port int `json:"port,string"` // string (!)
```
According to the documentation this might work (the conversion is done by the json lib) https://pkg.go.dev/encoding/json#Marshal

Or even convert only non-string fields as you did.
Then the whole thing is marshaled to `Migration` and it's done.

Issue #1149, opened 2026-04-22
Add support for references in KafkaSchema


albertw-dojo:
## Issue

At Dojo, we leverage the Aiven operator with Crossplane to provide a self-service platform for our developers. A key part of this is enabling developers to manage their Kafka message schemas. However, the current KafkaSchema CRD does not support schema references, which are essential for defining complex schemas in both Protobuf and JSON Schema that rely on external definitions. This limitation prevents us from automatically provisioning Kafka topics for schemas that are split across multiple files.

Currently the `KafkaSchema` specification does not support the ability to define [references](https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#schema-references) to other files. Proto files and JSON schemas both allow for importing external definitions which means they are not compatible with the current `KafkaSchema`.

We'd like to propose adding support to define the references on the CRD and pass them through to the [API calls made](https://github.com/aiven/aiven-operator/blob/8262d5cf78cc7e5d0c5d5f3488544133007cc5c9/controllers/kafkaschema_controller.go#L60) in the reconciliation loops. The generated API already supports the [references](https://github.com/aiven/go-client-codegen/blob/a5685a4be90f3d15bc65f982d5d29ce3d35aa66f/handler/kafkaschemaregistry/kafkaschemaregistry.go#L401) field required.

### 1. schemaReference field

Add new `schemaReference` array field with `name`, `subjectName`, `version` directly into the spec and pass them through to the API call. This approach judging from the existing code would be relatively easy to implement but has some drawbacks for users. Mainly the `version` has to be manually update on resources which is only known after making the API request.

*Example YAML*
```yaml
apiVersion: aiven.io/v1alpha1
kind: KafkaSchema
metadata:
  name: my-schema
spec:
  ...
  # Add schemaReferences to allow specifying the references needed in the API request
  schemaReference:
    - name: shared.proto
       subjectName: name.of.imported.proto.v1
       version: 1
```

Examples of [Confluent Crossplane Provider](https://marketplace.upbound.io/providers/crossplane-contrib/provider-confluent/v1.0.0/resources/confluent.crossplane.io/Schema/v1alpha1#doc:spec-forProvider-schemaReference) which has support for this feature and drawn inspiration from.

### 2. KafkaSchema Ref field

[...]

PR #1150, opened 2026-04-24
feat(kafkaschema): add schema references field for Protobuf/JSON

vmyroslav:

- Added references field to `KafkaSchema` spec, enabling `Protobuf` and `JSON` schemas that import other schemas

Resolves: NEX-2469.
Contributes to: #1149.


controllers/kafkaschema_controller.go
```diff
@@ -57,15 +57,29 @@ func (h KafkaSchemaHandler) createOrUpdate(ctx context.Context, avnGen avngen.Cl
...
+		refs := make([]kafkaschemaregistry.ReferenceIn, len(schema.Spec.References))
+		for i, r := range schema.Spec.References {
+			refs[i] = kafkaschemaregistry.ReferenceIn{
+				Name:    r.Name,
+				Subject: r.Subject,
+				Version: r.Version,
+			}
+		}
+		postIn.References = &refs
+	}
```
byashimov, line 75: Should it include an empty list for `References` when user removes them?

vmyroslav: No, the schema should be idempotent. So when user removes the `References`, we will send nil and it should create a new version of the schema.

byashimov (approved):

PR #1155, opened 2026-04-28
feat(mysql/postgres): added migrationSecretSource secret cleanup

vmyroslav:

- Add `deleteAfterMigration` flag. Opt-in, default false. Deletes the K8s Secret after `MigrationComplete=True`
- Skip re-resolving credentials and polling migration status once `Done`

Resolves: NEX-2473


api/v1alpha1/mysql_types.go
```diff
@@ -15,6 +15,8 @@ type MySQLSpec struct {
 	// Reference to a Secret containing migration credentials.
 	// Secret keys must match userConfig.migration JSON field names.
 	// If set, takes precedence over userConfig.migration.
+	// Leading and trailing whitespace is stripped from every value, including the password.
```
ndk: It looks like we're trying to automatically fix the user's typo. Shouldn't we validate the input instead?

api/v1alpha1/common.go
```diff
@@ -282,10 +282,20 @@ func ErrorSubstrChecker(substrings ...string) func(error) bool {
...
 	// +kubebuilder:validation:MinLength=1
 	// Name of the Secret containing migration credentials
 	Name string `json:"name"`
+
+	// +kubebuilder:default=false
+	// When true, the operator deletes the referenced Kubernetes Secret after
+	// migration completes successfully. Defaults to false.
+	// This field remains mutable after migration has completed: flipping it from
+	// false to true on a Done resource triggers Secret deletion on the next reconcile.
+	DeleteAfterMigration bool `json:"deleteAfterMigration,omitempty"`
```
ndk, line 298: What should happen if this field is flipped from true to false after the CR reaches Done?

api/v1alpha1/mysql_types.go
vmyroslav: Cause this keeps raising questions. I removed the striping and went with validation

api/v1alpha1/common.go
vmyroslav: Updated the documentation about the behaviour.

controllers/migration_secret.go
```diff
@@ -3,17 +3,31 @@ package controllers
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
 )
 
+// parseMigrationPort parses the "port" value from a migration secret.
+func parseMigrationPort(secretName, raw string) (int, error) {
```
ndk: Nit: `secretName` is only used here for error wording. Let's remove it from `parseMigrationPort` and add the Secret context at the call site instead.

controllers/migration_secret.go
```diff
@@ -22,7 +36,47 @@ func readMigrationSecret(ctx context.Context, k8s client.Reader, namespace strin
 
 	data := make(map[string]string, len(secret.Data))
 	for k, v := range secret.Data {
-		data[k] = strings.TrimSpace(string(v))
+		s := string(v)
+		if s != strings.TrimSpace(s) {
```
ndk, line 40: Nit: As far as I know, Postgre/MySQL don't forbid leading/trailing spaces in passwords. This is an operator-level policy, so if it's not documented already, let's make that explicit for users.

tests/migration_secret_test.go
ndk: This test looks outdated now.

tests/migration_secret_test.go
```diff
@@ -230,6 +231,129 @@ spec:
...
+	assert.True(t, apierrors.IsNotFound(err), "migration secret should be deleted, got err: %v", err)
+
+	// 7. Trigger a reconcile after deletion and verify the target stays healthy.
+	//    The skip-read guard in createOrUpdate relies on MigrationComplete=True to
+	//    avoid re-resolving the (now missing) Secret.
+	require.NoError(t, k8sClient.Get(ctx, types.NamespacedName{Name: targetName, Namespace: defaultNamespace}, targetPg))
+	if targetPg.Annotations == nil {
+		targetPg.Annotations = map[string]string{}
+	}
+	targetPg.Annotations["aiven.io/trigger-reconcile"] = randName("touch")
```
ndk: It seems this doesn't test what it should test. Updating only an annotation doesn't change `.metadata.generation`, so the reconciler may return early before `createOrUpdate`. Maybe update some safe spec field instead?

ndk (approved):

ndk (dismissed):

tests/migration_secret_test.go
vmyroslav: Updated the test

controllers/migration_secret.go
vmyroslav: There is an [explicit](https://github.com/aiven/aiven-operator/pull/1155/changes#diff-742d8c7992c347423f6d4a68abce1be8139f98c9186b7902b3bfb76b24386f7bR146) message in the description.

ndk (approved):

PR #1160, opened 2026-05-05
refactor(kafkaschema): migrate to managed reconciler

vmyroslav:

Migrates `KafkaSchema`to the new reconciler.
Resolves: NEX-2484

## Workaround for a reconciler race
- Diagnosing flaky e2e runs uncovered a pre-existing race in `controllers/reconciler.go`'s deferred `updateStatus`: a self-triggered follow-up reconcile pass can read `Status` from a stale cache and clobber the value the previous pass just persisted. `Status.ID` settles at the previous generation's value permanently, until the spec is edited again.

- **Workaround**: `applySchema` persists `Status.ID` synchronously via its own `r.Status().Update`

### DIFF
--- a/controllers/kafkaschema_controller.go
+++ b/controllers/kafkaschema_controller.go
@@ -8,55 +8,129 @@ import (
 
 	avngen "github.com/aiven/go-client-codegen"
 	"github.com/aiven/go-client-codegen/handler/kafkaschemaregistry"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
 )
 
-// KafkaSchemaReconciler reconciles a KafkaSchema object
-type KafkaSchemaReconciler struct {
-	Controller
-}
-
-func newKafkaSchemaReconciler(c Controller) reconcilerType {
-	return &KafkaSchemaReconciler{Controller: c}
-}
-
-type KafkaSchemaHandler struct{}
-
 //+kubebuilder:rbac:groups=aiven.io,resources=kafkaschemas,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=aiven.io,resources=kafkaschemas/status,verbs=get;update;patch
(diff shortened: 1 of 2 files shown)


controllers/kafkaschema_controller.go
```diff
@@ -8,212 +8,192 @@ import (
...
-func (h KafkaSchemaHandler) get(ctx context.Context, avnGen avngen.Client, obj client.Object) (*corev1.Secret, error) {
-	schema, err := h.convert(obj)
-	if err != nil {
-		return nil, err
+func (r *KafkaSchemaController) Create(ctx context.Context, schema *v1alpha1.KafkaSchema) (CreateResult, error) {
+	delete(schema.GetAnnotations(), instanceIsRunningAnnotation)
+	if err := r.applySchema(ctx, schema); err != nil {
+		return CreateResult{}, err
 	}
+	return CreateResult{}, nil
```
ndk, line 108: Let's do the same as we do in `controllers/kafkatopic_controller.go`. Otherwise, we may keep a stale `Running=True` condition after a schema update. Plus we we need to preserve the previous semantic.

```suggestion

	const reason = "CreatedOrUpdated"
	meta.SetStatusCondition(&topic.Status.Conditions, getInitializedCondition(reason, "Successfully created or updated the instance in Aiven"))
	meta.SetStatusCondition(&topic.Status.Conditions, getRunningCondition(metav1.ConditionUnknown, reason, "Successfully created or updated the instance in Aiven, status remains unknown"))

		return CreateResult{}, nil
```

controllers/kafkaschema_controller.go
```diff
@@ -8,212 +8,192 @@ import (
...
-
-	schema.Status.Version = version
-	meta.SetStatusCondition(&schema.Status.Conditions,
-		getRunningCondition(metav1.ConditionTrue, "CheckRunning",
-			"Instance is running on Aiven side"))
-
-	metav1.SetMetaDataAnnotation(&schema.ObjectMeta, instanceIsRunningAnnotation, "true")
-
-	return nil, nil
+	return UpdateResult{}, nil
```
ndk, line 121: Let's do the same as we do in `controllers/kafkatopic_controller.go`. Otherwise, we may keep a stale `Running=True` condition after a schema update. Plus we we need to preserve the previous semantic.

```suggestion

	const reason = "CreatedOrUpdated"
	meta.SetStatusCondition(&topic.Status.Conditions, getInitializedCondition(reason, "Successfully created or updated the instance in Aiven"))
	meta.SetStatusCondition(&topic.Status.Conditions, getRunningCondition(metav1.ConditionUnknown, reason, "Successfully created or updated the instance in Aiven, status remains unknown"))

		return UpdateResult{}, nil
```

ndk (approved):

PR #1163, opened 2026-05-07
fix(connectionpool): make databaseName and username immutable

ndk:
Resolves: NEX-2471

Prevent existing `ConnectionPool` resources from changing `spec.databaseName` or `spec.username`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Valkey major version
 - Add `MySQL` and `PostgreSQL` field `migrationSecretSource.deleteAfterMigration`, type `boolean`: When
   true, the operator deletes the referenced Secret after migration completes successfully. Defaults to `false`.
 - Remove the 63-character limit for the `KafkaSchema` field `subjectName`
+- Make `ConnectionPool` fields `databaseName` and `username` immutable
 
 ## v0.37.0 - 2026-04-09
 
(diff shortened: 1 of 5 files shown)


tests/connectionpool_test.go
```diff
@@ -18,13 +18,111 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
vmyroslav: This looks like a duplicate of the line above

tests/connectionpool_test.go
```diff
@@ -18,13 +18,111 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
 	"github.com/aiven/aiven-operator/controllers"
 )
 
+func TestConnectionPoolImmutableIdentityFields(t *testing.T) {
```
vmyroslav: nit: I'm not sure if we need an integration test just to check CEL validation.
If you really think that we need it, lets add pls `t.Parallel` and `s.Destroy(t)` to cleanup resources

vmyroslav (dismissed):

PR #1176, opened 2026-05-20
feat(kafka_schema): added kafka schema references

vmyroslav:

- Added `kafkaSchemaRef` to `KafkaSchema.Spec.References[]` so a dependent KafkaSchema can resolve `subject` and `version` from another KafkaSchema in the same namespace
- Dependents auto-propagate when the referent advances: a controller-runtime field index + watch + predicate re-enqueues dependents on referent `Status.Version` changes.
- Delete is now a two-step soft + hard delete on the registry subject. Re-applying a `KafkaSchema` with the same `subjectName` after deletion starts cleanly at version 1
- Self-reference (`A → A`) blocked at admission
- `Spec.References` now uses `+listType=map` keyed by `name`, with `MaxItems=100`. Reference name uniqueness is enforced by the API server.
- `kafkaSchemaVersionChangedPredicate` filters watch events to referent moves, avoiding noisy re-enqueues

Resolves: NEX-2501 & #1149


ndk (dismissed):

controllers/kafkaschema_controller.go
```diff
@@ -182,6 +274,67 @@ func (r *KafkaSchemaController) applySchema(ctx context.Context, schema *v1alpha
...
+		subject, version := ref.Subject, ref.Version
+		if ref.KafkaSchemaRef != nil {
+			target := &v1alpha1.KafkaSchema{}
+			key := client.ObjectKey{
+				Namespace: schema.GetNamespace(),
+				Name:      ref.KafkaSchemaRef.Name,
+			}
+			if err := r.Get(ctx, key, target); err != nil {
+				return nil, fmt.Errorf("resolving kafkaSchemaRef %s: %w", key, err)
+			}
```
byashimov, line 293: Can it be that the ref does not exist yet? I mean, should we avoid "not found" here?

byashimov (dismissed):

controllers/kafkaschema_controller.go
```diff
@@ -203,10 +356,66 @@ func (r *KafkaSchemaController) persistStatusID(ctx context.Context, schema *v1a
...
+		client.MatchingFields{kafkaSchemaRefIndex: schema.GetName()},
+	); err != nil {
+		return nil, err
+	}
+	names := make([]string, 0, len(list.Items))
+	for i := range list.Items {
+		if list.Items[i].GetName() == schema.GetName() {
+			continue
+		}
+		names = append(names, list.Items[i].GetName())
```
byashimov, line 418: Took me a bit it realise this is a self-ref protection.

controllers/kafkaschema_controller.go
vmyroslav: This case is handled upstream.
Reconcile calls `resolveK8sRefs` before `Observe`/`Update`. `resolveK8sRefs` does its own `Get` per `GetRefs()` entry, and requeues on `apierrors.IsNotFound`.
So `resolveReferences` only runs when every referent is already `Ready`.

controllers/kafkaschema_controller.go
vmyroslav: I'll work on the PR on top of this one and add breadcrumbs comment.

ndk (approved):

PR #1178, opened 2026-05-21
feat(upgradepipeline): add upgrade pipeline step support

ndk:
Resolves: NEX-2532

Add `UpgradePipelineStep` for upgrade pipeline steps.


controllers/upgradepipelinestep_controller.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if err := r.avnGen.UpgradePipelineStepDelete(ctx, cr.Spec.OrganizationID, cr.Status.ID); err != nil && !isNotFound(err) {
+		return fmt.Errorf("deleting upgrade pipeline step: %w", err)
+	}
+
+	return nil
+}
+
+func (r *UpgradePipelineStepController) lookupStep(ctx context.Context, cr *v1alpha1.UpgradePipelineStep) (*upgradepipeline.StepOut, error) {
+	out, err := r.avnGen.UpgradePipelineStepList(ctx, cr.Spec.OrganizationID)
```
vmyroslav: Based on the response type:
```go
type UpgradePipelineStepListOut struct {
	First      *string   `json:"first,omitempty"` // First page
	Last       *string   `json:"last,omitempty"`  // Last page
	Next       *string   `json:"next,omitempty"`  // Next page
	Prev       *string   `json:"prev,omitempty"`  // Previous page
	Steps      []StepOut `json:"steps"`
	TotalCount *int      `json:"total_count,omitempty"` // Total number of results
}
```

I assume that the pagination is possible, so we need to handle it.

controllers/upgradepipelinestep_controller.go
```diff
@@ -0,0 +1,176 @@
...
+		return UpdateResult{}, fmt.Errorf("updating upgrade pipeline step: %w", err)
+	}
+
+	r.applyStatus(cr, upgradepipeline.StepOut(*step))
+
+	return UpdateResult{}, nil
+}
+
+func (r *UpgradePipelineStepController) Delete(ctx context.Context, cr *v1alpha1.UpgradePipelineStep) error {
+	if cr.Status.ID == "" {
```
vmyroslav: I thought that we agreed to avoid dependency on the status in the flow

controllers/upgradepipelinestep_controller.go
```diff
@@ -0,0 +1,176 @@
...
+
+		existing = step
+	}
+
+	return existing, nil
+}
+
+func (*UpgradePipelineStepController) applyStatus(cr *v1alpha1.UpgradePipelineStep, step upgradepipeline.StepOut) {
+	cr.Status.ID = step.StepId
+	cr.Status.LastValidation = &v1alpha1.UpgradePipelineStepLastValidationStatus{
```
vmyroslav: I noticed in some of our CRD statuses we set many empty fields. Does it make sense to check that fields are not-empty? WDYT?

api/v1alpha1/upgradepipelinestep_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern="^[a-z][-a-z0-9]+$"
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// DestinationServiceName is the service name that waits for the source service.
+	DestinationServiceName string `json:"destinationServiceName"`
+
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=7
+	// AutoValidationDelayDays is the number of days before Aiven can automatically validate the source service.
+	AutoValidationDelayDays int `json:"autoValidationDelayDays,omitempty"`
```
vmyroslav: The field is declared as `int` with `omitempty` and default of 7. If user sets autoValidationDelayDays: 0, `JSON` marshaling drops the field, so we apply the default and the user's 0 becomes 7.

```go
func TestUpgradePipelineStepSpec_AutoValidationDelayDaysZero(t *testing.T) {
	spec := UpgradePipelineStepSpec{
		OrganizationID:          "org123",
		SourceProjectName:       "sandbox",
		SourceServiceName:       "src",
		DestinationProjectName:  "prod",
		DestinationServiceName:  "dst",
		AutoValidationDelayDays: 0,
	}

	raw, err := json.Marshal(spec)
	require.NoError(t, err)

	var decoded map[string]any
	require.NoError(t, json.Unmarshal(raw, &decoded))

	_, present := decoded["autoValidationDelayDays"]
	require.False(t, present)
}
```

controllers/upgradepipelinestep_controller.go
```diff
@@ -0,0 +1,179 @@
...
+		upgradepipeline.UpgradePipelineStepListSourceProjectName(cr.Spec.SourceProjectName),
+		upgradepipeline.UpgradePipelineStepListSourceServiceName(cr.Spec.SourceServiceName),
+		upgradepipeline.UpgradePipelineStepListDestinationProjectName(cr.Spec.DestinationProjectName),
+		upgradepipeline.UpgradePipelineStepListDestinationServiceName(cr.Spec.DestinationServiceName),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("listing upgrade pipeline steps: %w", err)
+	}
+
+	if (out.TotalCount != nil && *out.TotalCount > 1) || (out.Next != nil && *out.Next != "") || len(out.Steps) > 1 {
```
vmyroslav, line 143: This condition is quite complex for understanding. As I see the API exposes limit param to the query. Can we use it for the same purpose?

vmyroslav (approved):

PR #1181, opened 2026-05-25
fix(controllers): wait for connection secret before ready

ndk:
Resolves: NEX-2511

Fix service readiness so a resource isn't reported as ready until its connection Secret has been successfully published.

### DIFF
--- a/controllers/basic_controller.go
+++ b/controllers/basic_controller.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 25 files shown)


controllers/basic_controller.go
```diff
@@ -327,6 +328,32 @@ func hasPendingMigration(o v1alpha1.AivenManagedObject) bool {
 	return cond != nil && cond.Reason == v1alpha1.MigrationReasonInProgress
 }
 
+func (i *instanceReconcilerHelper) resourceNeedsConnectionSecretSync(ctx context.Context, o v1alpha1.AivenManagedObject) bool {
+	if GetIsRunningAnnotation(o) == "false" {
```
byashimov: I would make a helper that just would return a real boolean.

controllers/basic_controller.go
```diff
@@ -327,6 +328,32 @@ func hasPendingMigration(o v1alpha1.AivenManagedObject) bool {
...
+	secret := &corev1.Secret{}
+	if err := i.k8s.Get(ctx, types.NamespacedName{Name: connectionSecretName(withSecret), Namespace: withSecret.GetNamespace()}, secret); err != nil {
+		if !apierrors.IsNotFound(err) {
+			i.log.Info("unable to verify connection secret ownership", "error", err)
+		}
+		return true
+	}
+
+	ref := metav1.GetControllerOf(secret)
+	return ref == nil || o.GetUID() == "" || ref.UID != o.GetUID()
```
byashimov: Please add a bit of explanation what happens here.

controllers/basic_controller.go
```diff
@@ -547,51 +556,40 @@ func (i *instanceReconcilerHelper) createOrUpdateInstance(ctx context.Context, o
...
 
-	_, err = controllerutil.CreateOrUpdate(ctx, i.k8s, secret, func() error {
-		if secret.Data == nil {
-			secret.Data = make(map[string][]byte)
+	secret := &corev1.Secret{}
+	if err := i.k8s.Get(ctx, types.NamespacedName{Name: connectionSecretName(withSecret), Namespace: withSecret.GetNamespace()}, secret); err != nil {
+		if !apierrors.IsNotFound(err) {
+			i.log.Info("unable to verify connection secret ownership", "error", err)
 		}
+		return true
```
vmyroslav, line 586: nit: I'm okay, to accept this downside, but just wanna highlight that we may end up in wrong state of the service if K8S will hick up for some reason.
It would ruin the signature and readability, but maybe it makes sense to return this error

vmyroslav (dismissed):

controllers/basic_controller.go
ndk: Yeah, true. It might confuse ppl. I'd rather keep this path simple for now and rework it when this controller is migrated to the new reconciler.

byashimov (approved):

PR #1183, opened 2026-05-27
chore(tests): added unit tests to CI workflow

vmyroslav:

- added unit test workflow to run on each PR
- fixed few failing unit tests

Resolves: NEX-2562

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,20 @@
+name: unit-tests
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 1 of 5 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+      - uses: arduino/setup-task@v2
+        with:
+          version: 3.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
ndk, line 19: Do the unittests need it?

vmyroslav: Yes, otherwise the GH will limit the amount of requests to its API, we dont use this token internally.

ndk (approved):

PR #1184, opened 2026-05-28
fix(service_user): ignore empty password

vmyroslav:

ServiceUser CRs whose password was changed directly in the database (e.g. `ALTER USER`) got stuck, Aiven's API returns `password: ""` indefinitely, the operator wrapped that as `errPreconditionNotMet`, and the target K8s secret kept serving a stale value.

- `Observe` no longer treats empty password as an error. With `connInfoSecretSource` set, the declared value is pushed back via `Update` → `ResetCredentials` (heals in one cycle). Without a source, the empty value is published as-is.

Resolves: NEX-2559

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -76,6 +76,7 @@ nav:
           - installation/uninstalling.md
           - guides/token-management.md
           - guides/deletion-policy.md
+          - guides/serviceuser-password-management.md
           - controllers/reconciler.md
           - controllers/clickhouseuser.md
       - Resources: &crds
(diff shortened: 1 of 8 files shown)


byashimov (dismissed):

docs/mkdocs.yml
```diff
@@ -76,6 +76,7 @@ nav:
           - installation/uninstalling.md
           - guides/token-management.md
           - guides/deletion-policy.md
+          - guides/serviceuser-password-management.md
```
byashimov, line 79: I think this document does not exist.

vmyroslav: Works on my machine 😆

byashimov: Probably setting `strict` might help https://www.mkdocs.org/user-guide/configuration/#strict
But then you might face unrelated issues.

ndk (approved):

PR #1207, opened 2026-06-19
feat(kafka_native_acl): migrated to new reconciler

vmyroslav:

- migrated `kafka_native_acl` to new reconciler

Resolves: NEX-2630

### DIFF
--- a/controllers/kafkanativeacl_controller.go
+++ b/controllers/kafkanativeacl_controller.go
@@ -8,65 +8,59 @@ import (
 
 	avngen "github.com/aiven/go-client-codegen"
 	"github.com/aiven/go-client-codegen/handler/kafka"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
 )
 
-// KafkaNativeACLReconciler reconciles a KafkaNativeACL object
-type KafkaNativeACLReconciler struct {
-	Controller
+//+kubebuilder:rbac:groups=aiven.io,resources=kafkanativeacls,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=aiven.io,resources=kafkanativeacls/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=aiven.io,resources=kafkanativeacls/finalizers,verbs=get;create;update
+
+// KafkaNativeACLController reconciles a KafkaNativeACL object.
+type KafkaNativeACLController struct {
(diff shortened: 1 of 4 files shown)


controllers/kafkanativeacl_controller.go
```diff
@@ -8,66 +8,58 @@ import (
...
-
-	return checkServiceIsOperational(ctx, avnGen, acl.Spec.Project, acl.Spec.ServiceName)
-}
-
-// createOrUpdate creates or updates an instance on the Aiven side.
-func (h KafkaNativeACLHandler) createOrUpdate(ctx context.Context, avnGen avngen.Client, obj client.Object, _ []client.Object) error {
-	acl, err := h.convert(obj)
-	if err != nil {
-		return err
+	if acl.Status.ID == "" {
```
ndk, line 41: I think Observe() should look up the ACL by spec instead of relying on Status.ID. One match can be adopted, multiple matches should fail explicitly so we don't guess ownership or create another duplicate.

controllers/kafkanativeacl_controller.go
```diff
@@ -8,66 +8,58 @@ import (
...
+
+	// The spec is immutable, so an existing ACL is always up to date.
+	return Observation{
+		ResourceExists:   true,
+		ResourceUpToDate: hasLatestGeneration(acl),
+	}, nil
+}
+
+func (r *KafkaNativeACLController) Create(ctx context.Context, acl *v1alpha1.KafkaNativeACL) (CreateResult, error) {
 	in := &kafka.ServiceKafkaNativeAclAddIn{
```
ndk, line 65:
```suggestion
	delete(acl.GetAnnotations(), instanceIsRunningAnnotation)

	in := &kafka.ServiceKafkaNativeAclAddIn{
```

controllers/kafkanativeacl_controller.go
vmyroslav: I implemented content-matching for `KafkaACL` because its spec is mutable.

`KafkaNativeACL` is immutable (XValidation: self == oldSelf)
- Once Status.ID is set it maps to the same spec.
- Behavior regression - the native API allows exact-duplicate ACLs, so hard failure can be an issue for services that reconciles today
- IMHO, adoption for ACLs which are ambiguous by default is quite risky

controllers/kafkanativeacl_controller.go
vmyroslav: I can add it, but really dont see why

ndk: If the API fails the resource will be in a wrong state that can confuse users.

vmyroslav: I forgot about `markInstanceRunning` in `Observe`

ndk: Yeah, need to fix it in `Reconciler[T]` later

vmyroslav: Updated. Didn't apply suggestion, cause it will fail the commitlint 😢

ndk (approved):

PR #1210, opened 2026-06-22
refactor(project): migrate controller to the new reconciler

ndk:
Resolves: NEX-2642

### DIFF
--- a/controllers/project_controller.go
+++ b/controllers/project_controller.go
@@ -8,49 +8,171 @@ import (
 
 	avngen "github.com/aiven/go-client-codegen"
 	proj "github.com/aiven/go-client-codegen/handler/project"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


vmyroslav:
Based on the discussion that we had I would also add the deprecation notice to this resource. In this PR or separately, whatever works for u.
Also I would add the record to the changelog, cause it would effect the behaviour slightly.

vmyroslav (dismissed):

api/v1alpha1/project_types.go
```diff
@@ -90,6 +90,7 @@ type ProjectStatus struct {
 
 // Project is the Schema for the projects API.
 // Info "Exposes secret keys": `PROJECT_CA_CERT`
+// +kubebuilder:deprecatedversion:warning="The Project resource is deprecated. Existing Project resources continue to be reconciled during the deprecation period."
```
byashimov: I think, it is a bit unfair to deprecate things without a proper alternative.

byashimov (changes requested):

api/v1alpha1/project_types.go
ndk: I can just drop it out. @vmyroslav are you ok if I drop the last commit?

vmyroslav: sure

vmyroslav (approved):

byashimov (approved):

PR #1211, opened 2026-06-22
feat(kafka_schema_registry_acl): migrated to new reconciler

vmyroslav:

Resolves: NEX-2640


controllers/kafkaschemaregistryacl_controller.go
```diff
@@ -8,150 +8,126 @@ import (
...
-	if !ok {
-		return nil, fmt.Errorf("cannot convert object to KafkaSchemaRegistryACL")
+	for _, v := range list {
+		if v.Id != nil && *v.Id == acl.Status.ACLId {
+			return true, nil
+		}
 	}
 
-	return db, nil
+	return false, nil
```
ndk:
```suggestion
	return slices.ContainsFunc(list, func(v kafkaschemaregistry.AclOut) bool {return v.Id != nil && *v.Id == acl.Status.ACLId}), nil
```

ndk (approved):

PR #1218, opened 2026-06-30
feat(clickhouse_grant): migrated to new reconciler

vmyroslav:

- migrated `ClickHouseGrant` to the new reconciler
- this is a pure lift-and-shift onto the new reconciler - behavior is intentionally unchanged

Resolves: NEX-2664

### DIFF
--- a/controllers/clickhousegrant_controller.go
+++ b/controllers/clickhousegrant_controller.go
@@ -10,8 +10,6 @@ import (
 	avngen "github.com/aiven/go-client-codegen"
 	"github.com/hashicorp/go-multierror"
 	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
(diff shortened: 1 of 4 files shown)


controllers/clickhousegrant_controller.go
```diff
@@ -10,154 +10,113 @@ import (
...
 	itemsInSpec := collectFunc()
 	itemsInDB, err := queryFunc(ctx, avnGen, g.Spec.Project, g.Spec.ServiceName)
 	if err != nil {
-		return false, err
+		return err
 	}
 	missingItems := utils.CheckSliceContainment(itemsInSpec, itemsInDB)
 	if len(missingItems) > 0 {
 		err = fmt.Errorf(errorMsgFormat, missingItems)
 		meta.SetStatusCondition(&g.Status.Conditions, getErrorCondition(errConditionPreconditions, err))
```
ndk, line 116: Note that we don't clear the error on the next successful reconciliation.

controllers/clickhousegrant_controller.go
```diff
@@ -10,154 +10,113 @@ import (
...
-type ClickhouseGrantHandler struct{}
+func (r *ClickhouseGrantController) Observe(ctx context.Context, g *v1alpha1.ClickhouseGrant) (Observation, error) {
+	if _, err := getServiceIfOperational(ctx, r.avnGen, g.Spec.Project, g.Spec.ServiceName); err != nil {
+		return Observation{}, err
+	}
 
-//+kubebuilder:rbac:groups=aiven.io,resources=clickhousegrants,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=aiven.io,resources=clickhousegrants/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=aiven.io,resources=clickhousegrants/finalizers,verbs=get;create;update
+	exists := wasEverApplied(g)
```
ndk, line 45: We don't check for drifts on the API side. So, continous reconciliation won't work here. Is it on purpose?

controllers/clickhousegrant_controller.go
vmyroslav: Its true, but it should be fixed in general reconciler for all cases, so I plan to do it in a separate PR.

controllers/clickhousegrant_controller.go
vmyroslav: Yes, for this resource its intentional. We already had issues with internal changes in ClickHouse. I would like to avoid maintaining internal ClickHouse logic in operator.
We may discuss it as a separate new feature, however at the moment this intentional behaviour.

ndk (approved):

PR #1219, opened 2026-06-30
fix(reconciler): protect auth secret finalizers

ndk:
Resolves: NEX-2636


controllers/secret_finalizer_gc_controller.go
```diff
@@ -81,6 +76,45 @@ func (c *SecretFinalizerGCController) SetupWithManager(mgr ctrl.Manager, hasDefa
 	return builder.Complete(c)
 }
 
+func secretFinalizerGCEventPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return protectedSecretIsDeleting(e.Object) },
```
vmyroslav, line 83: nit: I would add a small comment hear, the cache replay is not the most obvious thing here, imho

vmyroslav (dismissed):

PR #1231, opened 2026-07-10
feat(database): migrate to new reconciler

vmyroslav:

- Migrated the `Database` controller to the `Reconciler[T]`
- `GetDatabaseByName` now **paginates** via the `Next` cursor instead of reading only the first page

Resolves: NEX-2663

### DIFF
--- a/controllers/database_controller.go
+++ b/controllers/database_controller.go
@@ -10,145 +10,128 @@ import (
 	"github.com/aiven/go-client-codegen/handler/service"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aiven/aiven-operator/api/v1alpha1"
 )
 
-// DatabaseReconciler reconciles a Database object
-type DatabaseReconciler struct {
-	Controller
-}
-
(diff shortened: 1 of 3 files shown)


controllers/database_controller.go
```diff
@@ -8,147 +8,124 @@ import (
...
-	if err != nil && !isNotFound(err) {
-		return false, err
+	if err := r.avnGen.ServiceDatabaseCreate(ctx, db.Spec.Project, db.Spec.ServiceName, &req); err != nil {
+		// The service can report RUNNING via ServiceGet while it is not ready
+		// to accept database creation yet, returning a transient 5xx.
+		if isServerError(err) {
+			return CreateResult{}, fmt.Errorf("%w: %w", errPreconditionNotMet, err)
+		}
+		return CreateResult{}, fmt.Errorf("cannot create database on Aiven side: %w", err)
 	}
```
ndk, line 75: I think we're missing the post-create transient state here. We can mirror the status condition updates in controllers/clickhousedatabase_controller.go. Otherwise, this migration regresses the existing behavior.

```suggestion

	const reason = "Created"
	meta.SetStatusCondition(&db.Status.Conditions, getInitializedCondition(reason, "Successfully created the instance in Aiven"))
	meta.SetStatusCondition(&db.Status.Conditions, getRunningCondition(metav1.ConditionUnknown, reason, "Successfully created the instance in Aiven, status remains unknown"))

```

vmyroslav: upd

ndk (approved):

PR #1234, opened 2026-07-16
feat(kafka_connector): migrate KafkaConnector to new reconciler

vmyroslav:

Resolves: NEX-2681

### DIFF
--- a/controllers/kafkaconnector_controller.go
+++ b/controllers/kafkaconnector_controller.go
@@ -5,8 +5,8 @@ package controllers
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
-	"strings"
 	"text/template"
 
 	avngen "github.com/aiven/go-client-codegen"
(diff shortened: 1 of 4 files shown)


controllers/kafkaconnector_controller.go
```diff
@@ -15,84 +15,167 @@ import (
...
 	}
 
 	// Sometimes GET (ServiceKafkaConnectGetConnectorStatus) returns OK,
 	// and POST (ServiceKafkaConnectCreateConnector) returns NotFound error.
 	// So instead of asking Aiven API if the connector exists,
-	// we try to create it.
-	_, err = avnGen.ServiceKafkaConnectCreateConnector(ctx, conn.Spec.Project, conn.Spec.ServiceName, &connCfg)
+	// we try to create it, and fall back to an edit on already-exists.
+	_, err = r.avnGen.ServiceKafkaConnectCreateConnector(ctx, conn.Spec.Project, conn.Spec.ServiceName, &connCfg)
 	if isAlreadyExists(err) {
```
ndk: I believe we should return the conflict and let the next reconciliation cycle observe the existing connector before updating it. Bypassing the Observe->Update flow risks racing with another actor.

vmyroslav: updated

ndk (approved):

PR #1235, opened 2026-07-16
upkeep(golang): update dependencies

x10an14-nav:
Especially controller-runtime.

## Update dependencies to latest, wherever that doesn't introduce any changes in yaml/k8s CRDs

Especially important/relevant for `controller-runtime`, which leads to transitive dependencies incompatibilities if one wants to import the go structs into their own, say, k8s operator. 

## Why this way

No other changes were introduced, AKA resulting YAML files evaluated identically when tested with a python script that imported them and compared datastructures.

PR #1238, opened 2026-07-17
feat: permit all Aiven API accepted user names for ServicsUser

x10an14-nav:
Currently, `aiven-operator` hardcodes the names of their `ServiceUser` CRs to be the same as the ServiceUser's name in the Aiven API.

However, K8s does not allow the same subset of strings for CR names as the Aiven API permits for ServiceUsers.

This change (while based on #1235, can be rebased off of it), fixes this by allowing a decoupling between the ServiceUser k8s CR name, and the Aiven ServiceUser it represents. 

## Why this way
It's a morally correct change, that also follows the pattern laid out by existing pattern laid out in multiple CRs declared by this project leveraging `connInfoSecretTarget`, e.g. see https://aiven.github.io/aiven-operator/resources/serviceuser.html#spec.connInfoSecretTarget for ServiceUser's own.


x10an14-nav:
Solved by https://github.com/aiven/aiven-operator/pull/1241

PR #1241, opened 2026-07-22
feat(service_user): added username field

vmyroslav:

Resolves: NEX-2751

### DIFF
--- a/api/v1alpha1/serviceuser_types.go
+++ b/api/v1alpha1/serviceuser_types.go
@@ -8,9 +8,23 @@ import (
 )
 
 // ServiceUserSpec defines the desired state of ServiceUser
+// +kubebuilder:validation:XValidation:rule="has(self.username) == has(oldSelf.username)",message="username can only be set during resource creation."
 type ServiceUserSpec struct {
 	ServiceDependant `json:",inline"`
 	SecretFields     `json:",inline"`
+
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="username is immutable."
+	// Username of the service user on Aiven.
+	// Defaults to the K8S resource name. Aiven accepts usernames that are not valid
+	// Kubernetes object names (e.g. containing underscores or uppercase characters);
+	// set this field to manage such users.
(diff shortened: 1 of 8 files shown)


api/v1alpha1/serviceuser_types.go
```diff
@@ -8,10 +8,24 @@ import (
 )
 
 // ServiceUserSpec defines the desired state of ServiceUser
+// +kubebuilder:validation:XValidation:rule="has(self.username) == has(oldSelf.username)",message="username can only be set during resource creation."
```
byashimov, line 11: Not sure, but this seems duplicates the "self == oldSelf" on the field, right?

vmyroslav: Not really, for our use case. As we add a new field to existing CRD.
`self == oldSelf` catches changing the value; adding the field to an existing resource passes through. The `has()` rule blocks this cases.

ndk (approved):

controllers/serviceuser_controller.go
```diff
@@ -152,19 +158,62 @@ func (r *ServiceUserController) Update(ctx context.Context, user *v1alpha1.Servi
...
+		sharedKey := client.ObjectKeyFromObject(shared).String()
+		logr.FromContextOrDiscard(ctx).Info(
+			"skipping deletion at Aiven: the same service user is managed by another ServiceUser resource",
+			"username", user.GetUsername(),
+			"sharedWith", sharedKey,
+		)
+		r.rec.Eventf(user, corev1.EventTypeWarning, eventSkippedDeletionAtAiven,
+			"skipped deleting user %q at Aiven: it is also managed by ServiceUser %s",
+			user.GetUsername(), sharedKey)
+		return nil
```
ndk, line 184: If `Delete` returns nil, `Reconciler` treats it as a successful remote deletion and emits `SuccessfullyDeletedAtAiven`, although the Aiven user was intentionally preserved. I see two options:
- Return an error handled specially by `Reconciler`
- The cleaner but broader one is to introduce `DeleteResult` so `Delete` can explicitly report whether the resource was deleted or preserved

vmyroslav: Thats a valid point. Here it covers very unlikely situation that should not occur under normal condition. But I agree, and we may separately and the `DeleteResult` as part of the interface.

PR #1248, opened 2026-07-29
feat(organizationproject): added OrganizationProject resource

vmyroslav:

- Adds a new `OrganizationProject` CRD and controller to manage Aiven projects that belong to an organization or organizational unit

Resolves: NEX-1471


api/v1alpha1/groupversion_info.go
```diff
@@ -47,6 +47,7 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&MySQL{}, &MySQLList{},
 		&OpenSearch{}, &OpenSearchList{},
 		&OpenSearchACLConfig{}, &OpenSearchACLConfigList{},
+		&OrganizationProject{}, &OrganizationProjectList{},
```
byashimov, line 50: Please update the `config/crd/kustomization.yaml` file, just to make work `make install`.

controllers/organizationproject_controller.go
```diff
@@ -0,0 +1,211 @@
...
+		return Observation{
+			ResourceExists:   true,
+			ResourceUpToDate: false,
+		}, nil
+	}
+
+	cert, err := r.avnGen.ProjectKmsGetCA(ctx, cr.Spec.ProjectID)
+	if err != nil {
+		return Observation{}, fmt.Errorf("getting project KMS CA: %w", err)
+	}
```
byashimov, line 82: I guess, when `NoSecret()` is false, this step can be skipped.

api/v1alpha1/organizationproject_types.go
```diff
@@ -0,0 +1,110 @@
...
+	// BasePort is the valid port number range for the project, from 10000 to 30000.
+	// When omitted, the field is unmanaged: Aiven assigns the value and changes made
+	// outside Kubernetes are left as is.
+	BasePort *int `json:"basePort,omitempty"`
+
+	// +kubebuilder:validation:MaxItems=10
+	// TechnicalEmails are the technical contact emails of the project.
+	// This list is authoritative: when omitted, emails added outside
+	// Kubernetes are removed.
+	TechnicalEmails []string `json:"technicalEmails,omitempty"`
```
byashimov, line 48: I think, this one needs additional validation that does not allow duplicates. Otherwise, the backend normailzation would produce infinitive loop in observe comparison.

controllers/organizationproject_controller.go
```diff
@@ -0,0 +1,211 @@
...
+
+	techEmails := organizationProjectTechEmails(cr.Spec.TechnicalEmails)
+	in := &organizationprojects.OrganizationProjectsCreateIn{
+		ProjectId:      cr.Spec.ProjectID,
+		BillingGroupId: cr.Spec.BillingGroupID,
+		ParentId:       NilIfZero(parentID),
+		BasePort:       cr.Spec.BasePort,
+		Tags:           emptyIfNil(cr.Spec.Tags),
+		TechEmails:     &techEmails,
+	}
```
byashimov, line 109: According to the documentation, and the docstring, the `parentID` is either "Organization or unit ID to where to move the project". So do we really need to resolve it back to the accountID or can we just use the value, user has set?

byashimov (changes requested):

ndk (commented):
I think you should add the new resource to config/crd/kustomization.yaml as well

controllers/organizationproject_controller.go
```diff
@@ -0,0 +1,211 @@
...
+
+	gotEmails := make([]string, len(got.TechEmails))
+	for i, e := range got.TechEmails {
+		gotEmails[i] = e.Email
+	}
+	if !cmp.Equal(cr.Spec.TechnicalEmails, gotEmails, cmpopts.SortSlices(strings.Compare), cmpopts.EquateEmpty()) {
+		return false, nil
+	}
+
+	parentID, err := r.resolveParentID(ctx, cr.Spec.ParentID)
```
ndk, line 189: Do we really need to resolve it if cr.Spec.ParentID == got.ParentId?

api/v1alpha1/groupversion_info.go
vmyroslav: Done.

controllers/organizationproject_controller.go
vmyroslav: I decided to copy the existing behaviour that we have in Terrafrom

api/v1alpha1/organizationproject_types.go
vmyroslav: Added normalization func that would prevent duplicates. The CEL validation is not enough to cover some of the cases

controllers/organizationproject_controller.go
byashimov: It's quite a legacy. Though I get your point.

controllers/organizationproject_controller.go
vmyroslav: The resolve function sorted in the cases that it wouldn't make a redundant calls if they are equal.

ndk: But how? `resolveParentID` gets only `parentID` from spec. It knows nothing about `got`

vmyroslav: Yes, `resolveParentID` would do a quick return the same `string` due to the condition:
```go
	if !strings.HasPrefix(parentID, "org") {
		return parentID, nil
	}
```
and 2 lines bellow we will: `return parentID == got.ParentId, nil`

ndk: Is it possible to have spec.ParentID = "org123" && got.ParentId = "org123"?

vmyroslav: No, I [added](https://github.com/aiven/aiven-operator/pull/1248/changes/ec03482da510f2644f6cac4a7a09e3711aa8a77d) small subtest to show it.

ndk (approved):

controllers/organizationproject_controller.go
vmyroslav: updated as well

byashimov (approved):

PR #1254, opened 2026-08-06
fix(kafkanativeacl,kafkaschemaregistryacl): adopt existing ACLs on CR creation

yungjakey:

`Observe` decided existence from `Status.ID`/`Status.ACLId` alone. The operator only writes that field on its own first create, so it is empty for any ACL it did not create, and empty was read as "does not exist".

- `KafkaNativeACL` then loops indefinitely on `409 Identical ACL entry already exists`.
- `KafkaSchemaRegistryACL` silently creates a duplicate, because `ServiceSchemaRegistryAclAdd` returns no 409. The original entry stays unmanaged.

Reachable two ways: 
- a CR deleted under `deletionPolicy: Orphan` and re-applied, or 
- a first apply of an ACL that already exists because it was created in the Console or by Terraform.

`Observe` now lists the remote ACLs when the status ID is empty and adopts an entry matching every immutable spec field, before falling back to `Create`.

Separately, `KafkaNativeACL` `Observe` returned `ResourceUpToDate: hasLatestGeneration(acl)` directly beneath a comment saying an existing ACL is always up to date, because the spec is immutable. 
It now returns `true`. 
The old form dispatched a no-op `Update` whenever `Status.ID` was set without the processed-generation annotation, which is what `kubectl patch --subresource=status` leaves behind.

## Why this way

Adoption marks the instance running and returns `ResourceUpToDate: true`, so it skips `Create` and `Update`. The spec is immutable, so `Update` has nothing to apply, and the running condition means only that the ACL is present on Aiven, which the list call just established.

`nativeSpecMatches` and `schemaRegistrySpecMatches` have table tests rather than reconciler coverage, because a missing field comparison does not fail any reconciler test. Adopting on a partial match would take ownership of an unrelated ACL and delete it with the CR.


yungjakey:
@phr0 @titaniumcoder

vmyroslav:
Thanks for this PR @yungjakey.

One thing we want to take a bit of time on: this is a behaviour change rather than a pure fix. Adopting an ACL the operator didn't create also means the operator will delete it when the custom resource goes away, and there are entries on a service that can be created by other things (the Console, Terraform, etc.), so we'd like to think through the consequences.

Sorry for the delay on review. Most likely we'll merge this roughly as-is and add our changes on top rather than asking you to iterate — we'll ping you, so you can see what changed and why.

yungjakey:
> it when the custom resource goes away, and there are entries on a service that can be created by other things (the Console, Terraform, etc.), so we'd like to think through the consequences.

got it, thanks for the update!

vmyroslav:
Thanks again @yungjakey 
Merging this now, as promised. Our follow-up changes will go up as a separate PR shortly and I'll link it here, so you can see what we adjusted and why.

PR #1255, opened 2026-08-06
fix(kafkaacl,kafkaquota,kafkaschemaregistryacl): reach IsReadyToUse in a single reconcile cycle

yungjakey:

`IsReadyToUse` requires the `instance-is-running` annotation and the processed-generation annotation. The reconciler writes processed-generation in `completeReconcileSuccess`, but the running annotation is the controller's job, and `Create` and `Update` never set it here. So the first cycle finished not ready and requeued after `requeueTimeout` (10 s), and only the next cycle's `Observe` marked it running. Anything gating on readiness through `GetRefs` waited that extra cycle.

All three now call `markInstanceRunning` before returning:

- `KafkaACL`: `Create` and `Update`.
- `KafkaQuota`: `Create` and `Update`, replacing the `Running=Unknown` write that the running state now contradicts.
- `KafkaSchemaRegistryACL`: `Create` only, leaving its adoption behaviour alone. `Update` is unreachable, because `Observe` never reports an existing ACL as out of date.

`KafkaNativeACL`, `ConnectionPool` and `UpgradePipelineStep` already did this.

## Why this way

For these kinds `Observe` already sets the running condition on nothing more than the resource being present on Aiven, so `Create` asserts what the next cycle would. Unlike `KafkaTopic` there is no provisioning state to wait on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
 - Add kind: `OrganizationProject` to manage Aiven projects that belong to an organization or organizational unit.
+- Fix `KafkaACL`, `KafkaQuota`, and `KafkaSchemaRegistryACL` to reach the `ReadyToUse`
+  state in a single reconcile cycle after creation or update.
 
 ## v0.43.0 - 2026-07-24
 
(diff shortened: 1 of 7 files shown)


yungjakey:
@phr0 @titaniumcoder

ndk (approved):

vmyroslav (approved):

PR #1265, opened 2026-08-18
fix(kafkanativeacl,kafkaschemaregistryacl): decide ACL existence by spec match

vmyroslav:

Follow-up to #1254, which adopted a matching `ACLs` only when the status ID was empty.

- Observe now decides existence by matching the spec against the live ACL list, on every reconcile.                                                                                                                                          
- **Behaviour change**: an ACL recreated outside the operator is taken over instead of left in an error state, and deleting the resource deletes the taken-over entry unless `deletionPolicy: Orphan` is set.
- An empty host from Aiven is treated as the "*" the spec defaults to.

Resolves: NEX-2794

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@
   present on Aiven instead of trying to create a second one. Existence is decided by matching
   every identifying field against the ACL list. This
   covers an ACLs created outside the operator as well. Note
-  that an adopted ACL is deleted together with the custom resource unless `deletionPolicy: Orphan`
-  is set.
+  that an adopted ACL is deleted together with the custom resource unless the
+  `controllers.aiven.io/deletion-policy: Orphan` annotation is set.
 - Fix `KafkaSchema` never converging when `schema` and `compatibilityLevel` change in the same apply:
   the compatibility level is now set before the new schema version is registered. Behavior change: a
   rejected schema leaves the new level applied, and tightening the level alongside a schema that violates it now fails.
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
 - `ServiceUser`: increased the amount of concurrent reconcilers up to 10
+- Fix `KafkaNativeACL` and `KafkaSchemaRegistryACL` to adopt the matching ACL that is already
+  present on Aiven instead of trying to create a second one. Existence is decided by matching
+  every identifying field against the ACL list. This
+  covers an ACLs created outside the operator as well. Note
+  that an adopted ACL is deleted together with the custom resource unless `deletionPolicy: Orphan`
```
byashimov:
```suggestion
  that an adopted ACL is deleted together with the custom resource unless `controllers.aiven.io/deletion-policy: Orphan`
```
Users might copy this, so let's have it the full name, or link it to the docs.

controllers/kafkanativeacl_controller.go
byashimov, line 107: Would it make sense to iterate over the name here too? I mean, the Observe adopts whatever it finds to adopt already.

byashimov (dismissed):

CHANGELOG.md
vmyroslav: Updated

byashimov (approved):

controllers/kafkanativeacl_controller.go
vmyroslav: I considered that, but I'd rather keep `Delete` on the cached **ID**. Deleting by spec match makes `Delete` destructive toward entries this CR never created or adopted, e.g. a CR that never successfully reconciled would still delete a matching ACL someone created externally.

In the current case we only delete what this CR verifiably tracked.

PR #1267, opened 2026-08-19
fix(kafkaschema): set compatibility level before registering schema

### LINKED ISSUE #1263: KafkaSchema: schema registered before compat level, so loose+breaking change never converges
A `KafkaSchema` never converges when `schema` and `compatibilityLevel` change in the same apply and the new schema is only valid under the new level.

[`applySchema`][applySchema] registers the schema version before setting the subject's compatibility level: the version [`POST`][post] runs first, then the config [`PUT`][put] is skipped unless the field is set. The registry validates a new version against the level in effect at POST time ([`get_compatibility_mode`][compat_mode], called from [`write_new_schema_local`][write_local]), so the new schema is checked against the *old* level -- the level the apply is about to replace.

Two failure modes:

1. **Loosening, e.g. `BACKWARD` -> `NONE`, to admit a breaking change.** The `POST` is rejected against the old `BACKWARD` level. `applySchema` returns early, the fingerprint annotation is never written, so `Observe` reports drift forever and the resource requeues in a permanent error loop.
2. **Tightening, e.g. `NONE` -> `FULL`, alongside a schema that violates it.** The `POST` succeeds against the old loose level, then the level is tightened and never re-validated, so an incompatible schema sits as `latest` while the resource reports `Running`.

The original ordering dates to the first KafkaSchema support ([`85f9ed3b`][orig]), where the `PUT` was guarded by `// set compatibility level if defined for a newly created Kafka Schema Subject`, implying a subject must exist before it can be configured. It does not: Karapace's [`config_subject_set`][config_set] emits a config message without any subject lookup. The managed-reconciler refactor ([`cd42055c`][refactor]) dropped the comment but kept the POST-then-PUT order.

Setting the level first mirrors Confluent Schema Registry semantics: `PUT /config/{subject}` creates the override without the subject existing, and registration validates against the subject level in effect at POST time ([API reference][confluent]).

[applySchema]: https://github.com/yungjakey/aiven-operator/blob/ebcd0c36/controllers/kafkaschema_controller.go#L216
[post]: https://github.com/yungjakey/aiven-operator/blob/ebcd0c36/controllers/kafkaschema_controller.go#L232
[put]: https://github.com/yungjakey/aiven-operator/blob/ebcd0c36/controllers/kafkaschema_controller.go#L251
[orig]: https://github.com/aiven/aiven-operator/commit/85f9ed3b
[refactor]: https://github.com/aiven/aiven-operator/commit/cd42055c
[config_set]: https://github.com/Aiven-Open/karapace/blob/680afdd9e4e3d140cd8101b634e5cd1b0fa98cc1/src/karapace/api/controller.py#L355
[compat_mode]: https://github.com/Aiven-Open/karapace/blob/680afdd9e4e3d140cd8101b634e5cd1b0fa98cc1/src/karapace/core/schema_registry.py#L118
[write_local]: https://github.com/Aiven-Open/karapace/blob/680afdd9e4e3d140cd8101b634e5cd1b0fa98cc1/src/karapace/core/schema_registry.py#L361
[confluent]: https://docs.confluent.io/platform/current/schema-registry/develop/api.html

yungjakey:

Reorders `applySchema` to set the subject's compatibility level before registering the new schema version, so the registry validates the version against the level the apply is about to install.

Resolves: #1263

Behavior change:
- Loosening the level alongside a schema that is only valid under it now succeeds instead of looping forever.
- Tightening the level alongside a schema that violates it now fails instead of silently succeeding.
- A schema the registry rejects leaves the new compatibility level applied and is retried on the next reconcile.

## Why this way

Setting the level first mirrors Confluent Schema Registry semantics: `PUT /config/{subject}` creates the override without the subject having any schema versions, and registration validates against the level in effect at POST time. Karapace's `config_subject_set` emits a config message without any subject lookup, so a subject holding only a config entry still 404s on version list and `Observe` treats it as absent -- the create path is unaffected.

`compatibilityLevel` stays in the same drift fingerprint as the schema body, so both land in a single `applySchema` pass; the write is unconditional because no sub-resource controller reads back its own config first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
 - `ServiceUser`: increased the amount of concurrent reconcilers up to 10
+- Fix `KafkaSchema` never converging when `schema` and `compatibilityLevel` change in the same apply:
+  the compatibility level is now set on the subject before the new schema version is registered, so a
+  schema that only becomes valid after a loosening change (for example `BACKWARD` to `NONE`) is accepted.
+  Behavior change: a schema the registry rejects now leaves the new compatibility level applied, and
+  tightening the level together with a schema that violates it now fails instead of silently succeeding
 
 ## v0.44.0 - 2026-08-11
 
--- a/api/v1alpha1/kafkaschema_types.go
+++ b/api/v1alpha1/kafkaschema_types.go
@@ -97,6 +97,10 @@ type KafkaSchemaStatus struct {
 // Deletion: the operator performs a soft delete followed by a hard delete on
 // the subject. The subject disappears from the registry's listing, re-applying a KafkaSchema with the same subjectName
 // after deletion starts a brand-new subject at version 1.
+//
+// Update ordering: when schema and compatibilityLevel change in the same apply, the
+// compatibility level is set first, because the registry validates a new version against the
+// level currently configured for the subject. A rejected schema leaves the new level applied.
 // +kubebuilder:printcolumn:name="Service Name",type="string",JSONPath=".spec.serviceName"
 // +kubebuilder:printcolumn:name="Project",type="string",JSONPath=".spec.project"
 // +kubebuilder:printcolumn:name="Subject",type="string",JSONPath=".spec.subjectName"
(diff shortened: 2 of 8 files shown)


vmyroslav:
Thanks a lot for this @yungjakey !

Issue #1261, opened 2026-08-13
`aiven-operator` `ServiceUser` reconcile latency


x10an14-nav:
# `aiven-operator`: `ServiceUser` reconcile latency grows linearly with CR count

Measured 2026-08-13 against [`aiven/aiven-operator` `v0.43.0`](https://github.com/aiven/aiven-operator/tree/v0.43.0).

## Problem

In a cluster with ~1,080 `ServiceUser` resources, a newly created `ServiceUser` waits a **median of 14–19 minutes** before the operator's first reconcile of it.
Connection secrets are only published after that first reconcile, so consumers wait the same 14–19 minutes for credentials.

## Measurements

Method: per CR, `min(status.conditions[].lastTransitionTime) − metadata.creationTimestamp`, grouped by creation day.
Values are upper bounds where a condition re-transitioned later; they still separate seconds from minutes.

| `ServiceUser` count in cluster | pickup latency (median) |
|---|---|
| ~60–120 | 0.0–0.6 min |
| ~900 | 14.0–14.7 min |
| ~1,065 | 18.1–18.75 min |

The ~900-CR rows include single CRs created on otherwise idle weekend days, so burst load is excluded: the standing population alone sets the latency.
Per-item cost is consistent across both loaded eras: 900 CRs/14.0 min ≈ 0.93 s, 1,065/18.75 min ≈ 1.06 s.

## Root cause

- The `serviceuser` controller runs with controller-runtime's default of **1 concurrent worker**; the only controller that raises it is `kafkatopic` ([`controllers/kafkatopic_controller.go#L21`](https://github.com/aiven/aiven-operator/blob/v0.43.0/controllers/kafkatopic_controller.go#L21)).
- Every `Ready` `ServiceUser` is re-queued every `PollInterval` ([`controllers/reconciler.go#L423`](https://github.com/aiven/aiven-operator/blob/v0.43.0/controllers/reconciler.go#L423)), which defaults to **10 minutes** ([`controllers/setup.go#L19`](https://github.com/aiven/aiven-operator/blob/v0.43.0/controllers/setup.go#L19)) and is not wired to any flag or environment variable.
- Each reconcile — including of an unchanged `Ready` user — makes at least two sequential Aiven API calls: `ServiceGet(..., IncludeSecrets)` ([`controllers/common.go#L107`](https://github.com/aiven/aiven-operator/blob/v0.43.0/controllers/common.go#L107)) plus `ServiceUserGet`, with no caching.

[...]

PR #1271, opened 2026-08-24
feat(reconciler): exposed pollInterval param

vmyroslav:

- Added a -`-poll-interval` CLI flag and a matching `pollInterval` Helm value to configure how often resources in a steady state are re-reconciled against the Aiven API. The interval was previously hardcoded to **10** minutes. 

Resolves: NEX-2813, Contributes to: #1261

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"go.uber.org/zap/zapcore"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -54,7 +55,11 @@ func main() {
 	var probeAddr string
 	var development bool
 	var webhookPort int
+	var pollInterval time.Duration
 	flag.IntVar(&webhookPort, "webhook-port", webhookDefaultPort, "Webhook server port (default: 9443)")
+	flag.DurationVar(&pollInterval, "poll-interval", 0,
```
ndk: If I'm not mistaken we can put `defaultPollInterval` here.

main.go
```diff
@@ -79,6 +84,12 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	// Validated before the manager is built, so a bad value costs nothing.
+	if err := controllers.ValidatePollInterval(pollInterval); err != nil {
```
ndk, line 88: I would move it into `SetupControllers` closer to the place where it's used.

ndk (dismissed):

main.go
vmyroslav: I put it in `main` intentionally: the range is a constraint on the flag, not on `SetupControllers`.
Tests call `SetupControllers` too and may want short intervals. `normalize()` already handles bad values there.

main.go
vmyroslav: done

ndk (approved):

PR #1277, opened 2026-08-27
feature(valkey): expose version

x10an14-nav:
## Why this way
Minimal approach found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
+- Add `Valkey` field `version`, type `string`: Valkey major version, an alias for `userConfig.valkey_version`
+
 ## v0.45.0 - 2026-08-24
 
 - `ServiceUser`: increased the amount of concurrent reconcilers up to 10
--- a/api/v1alpha1/valkey_types.go
+++ b/api/v1alpha1/valkey_types.go
@@ -12,8 +12,12 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // ValkeySpec defines the desired state of Valkey
+// +kubebuilder:validation:XValidation:rule="!(has(self.version) && has(self.userConfig) && has(self.userConfig.valkey_version))",message="cannot use both version and userConfig.valkey_version"
 type ValkeySpec struct {
 	ServiceCommonSpec `json:",inline"`
+
+	// Valkey major version, an alias for `userConfig.valkey_version`
+	Version string `json:"version,omitempty"`
 
 	// Valkey specific user configuration options
 	UserConfig *valkeyuserconfig.ValkeyUserConfig `json:"userConfig,omitempty"`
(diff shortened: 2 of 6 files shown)


x10an14-nav:
Cross-reference w/Aiven support ticket: https://support.aiven.io/hc/en-us/requests/107334

vmyroslav:
Thanks for the PR @x10an14-nav.
Just to make sure I've understood correctly — your goal is to be able to read a service's version from the resource even when it wasn't set in the CRD, correct?

If yes:
- We can't take this approach though, because it doesn't do what you asked for. You said you need a `version` field you can read, not set. If you don't set it, it stays empty, so it can't tell you what version a service is actually running.
- On the OpenSearch comparison: we can't find a version field on any service CRD, OpenSearch included. If you're seeing it somewhere on the CRD, a snippet would help.
- We can find another way to achieve this.

x10an14-nav:
> Just to make sure I've understood correctly — your goal is to be able to read a service's version from the resource even when it wasn't set in the CRD, correct?
> (...)
> We can't take this approach though, because it doesn't do what you asked for. You said you need a version field you can read, not set. If you don't set it, it stays empty, so it can't tell you what version a service is actually running.

Correct, because it will be set in the Aiven API for the resource in question.
(Upon creation of a new one we don't really care, and are happy with the default the Aiven API implements).

> On the OpenSearch comparison: we can't find a version field on any service CRD, OpenSearch included. If you're seeing it somewhere on the CRD, a snippet would help.

Hmm, seems I might be at fault here, trusting some 3rd-party without vetting myself.
Seems upon further investigation that there's caching on our side from the aiven API.

> We can find another way to achieve this.

Sure, I don't think this is that technically difficult/unreasonable.
The only drawback I can think of is how it relates to your `userConfig`.

vmyroslav:
Since the goal is to read the version a service is actually running, we'd propose solving this with a `status.version` field [here](https://github.com/aiven/aiven-operator/blob/main/api/v1alpha1/common.go#L59), instead of a spec-side.  Spec is desired state: it stays empty unless you set it.

Would it work for you?

x10an14-nav:
> Since the goal is to read the version a service is actually running, we'd propose solving this with a `status.version` field [here](https://github.com/aiven/aiven-operator/blob/main/api/v1alpha1/common.go#L59), instead of a spec-side. Spec is desired state: it stays empty unless you set it.
> 
> Would it work for you?

Sounds weird to me to separate "desired" from "actual" with 2x different locations in CRD, but sure, this solution you suggest would work for us! =)

vmyroslav:
Glad it works for you.                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                     
It's not something specific to this operator: same as `spec.replicas` vs `status.readyReplicas` on a `Deployment`.
It also avoids the duplication concern from the original approach: a `spec.version` next to the existing `userConfig.valkey_version` would mean two writable fields for the same setting.

We'll implement this on our side and expose `status.version`. We'll link the PR here once it's ready — thanks again for raising this.

vmyroslav:
As discussed before, this PR is closed in a favour of #1280

vmyroslav:
The new version [v0.46.0](https://github.com/aiven/aiven-operator/releases/tag/v0.46.0) is released.

x10an14-nav:
Thank you! =)

PR #1289, opened 2026-09-15
fix(serviceuser): pass authentication when creating users

ndk:
Resolves: NEX-2846

Pass `spec.authentication` to the Aiven API when creating a `ServiceUser`, so the requested MySQL authentication method is respected..


vmyroslav:
There is one concern for me: any `ServiceUser` created before this fix with authentication set will never converge, because only `Create` sends the field.
We may examine an option to do this within update as well during the `ServiceUserCredentialsModify` call OR document the limitation and explicitly set the immutability rule for this field.

controllers/serviceuser_controller.go
```diff
@@ -144,7 +147,23 @@ func (r *ServiceUserController) Update(ctx context.Context, user *v1alpha1.Servi
 		logr.FromContextOrDiscard(ctx).V(1).Info("skipping access control update since it isn't provided in the spec")
 	}
 
-	wrotePassword, err := r.setAivenPasswordIfProvided(ctx, user, password)
+	authentication := user.Spec.Authentication
+	if authentication != "" {
```
vmyroslav, line 151: nit: I would add the description of this behaviour to the field itself as it may be not completely expected. Also we can emit an event when we see that auth spec changed

PR #1290, opened 2026-09-15
feat(serviceuserrotation): add scheduled credential rotation

ndk:
Resolves: NEX-2675.

Changing a service user's password can break applications that haven't loaded the new password yet. This PR adds `ServiceUserRotation` to rotate credentials on a schedule while giving applications time to switch.

Users provide exactly two existing Aiven service users, a rotation interval of at least `1h` and a target Kubernetes Secret. Applications use this Secret as their stable source of credentials.

The first publication changes the first user's password immediately. With users `a` and `b`, switching the Secret to `b` leaves `a`'s password unchanged until the next rotation back to `a`. Applications normally have one rotation interval to reload the credentials from the Secret or restart.

Users and permissions are managed separately through `ServiceUser` or other tools. When using `ServiceUser`, it must leave password management to rotation and disable its own connection Secret.

Rotation has to update both Aiven and Kubernetes. If the Aiven password change succeeds but the Secret update fails, retries must finish publishing the same credentials. This is the main case to focus on during review.

Rotation assumes exclusive control of password changes. It doesn't restore passwords changed outside rotation, since restoring an old password could break clients using the new one.

Deleting `ServiceUserRotation` stops rotation and removes its Secret. Aiven users, passwords and permissions stay unchanged.

The [rotation guide](docs/docs/guides/serviceuser-rotation.md) describes setup and recovery.


api/v1alpha1/serviceuserrotation_types.go
```diff
@@ -0,0 +1,113 @@
...
+	// The order determines the rotation sequence. The controller creates missing users and adopts matching existing users.
+	// Unless the Orphan deletion policy is used, deleting this resource deletes every user in the pool.
+	// +kubebuilder:validation:MinItems=2
+	// +kubebuilder:validation:MaxItems=100
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=64
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, y == x))",message="usernames must be unique."
+	// +kubebuilder:validation:XValidation:rule="self.all(username, username != 'avnadmin')",message="avnadmin cannot be managed by ServiceUserRotation."
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="usernames is immutable."
+	Usernames []string `json:"usernames"`
```
vmyroslav, line 26: TBH, I can't see a lot of benefits from the list, but it creates some confusion, IMHO. WDYT about limiting it to 2 users only?

api/v1alpha1/serviceuserrotation_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// RotationInterval is the minimum time between scheduled credential rotations.
+	// Measured from the publication time saved in the connection Secret; must be at least one minute.
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1m')",message="rotationInterval must be at least 1m."
+	RotationInterval metav1.Duration `json:"rotationInterval"`
```
vmyroslav, line 31: I would say that 1m is too low threshold for it. We have default reconcile period for 10 minutes.

controllers/serviceuserrotation_controller.go
```diff
@@ -0,0 +1,406 @@
...
+	for key, value := range details {
+		secret.Data[key] = []byte(value)
+	}
+	target := cr.GetConnInfoSecretTarget()
+	secret.Labels = maps.Clone(target.Labels)
+	secret.Annotations = maps.Clone(target.Annotations)
+	return nil
+}
+
+func (r *ServiceUserRotationReconciler) ensureUsers(ctx context.Context, cr *v1alpha1.ServiceUserRotation, cl avngen.Client, svc *service.ServiceGetOut) error {
```
vmyroslav: The adoption scenario looks quite risky to me. You can delete existing `ServiceUser` just by duplicating the name.
Deleting the spec leads to deletion of all `ServiceUsers`. Maybe it makes sense to enable orphan be default here or add another `deletion-protection` annotation.

controllers/serviceuserrotation_controller.go
```diff
@@ -0,0 +1,406 @@
...
+		return err
+	}
+
+	cl, err := r.aivenClient(ctx, cr)
+	if err != nil {
+		return err
+	}
+	svc, err := getServiceIfOperational(ctx, cl, cr.Spec.Project, cr.Spec.ServiceName)
+	if err != nil {
+		return err
```
vmyroslav, line 129: I guess we may wrap this error here if the service is not ready yet, in case we applied multiple specs together.

api/v1alpha1/serviceuserrotation_types.go
byashimov: Agreed on 1h 👍

api/v1alpha1/serviceuserrotation_types.go
byashimov: Agreed on max=2 👍

controllers/serviceuserrotation_controller.go
byashimov: Agreed on letting ServiceUser CR to manage users. This CR is to manage passwords only.

CHANGELOG.md
```diff
@@ -3,8 +3,10 @@
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
 - Add up to 10% jitter to the periodic reconcile interval.
+- Add kind: `ServiceUserRotation` to manage a rotating pool of service users behind a stable connection Secret.
 - Change `Kafka` field `userConfig.karapace_version`: pattern ~~`^[0-9]+\.[0-9]+\.[0-9]+$`~~
 - Remove the character pattern from `KafkaTopic` `tags` key and value; only the length limits remain.
+- Change `ServiceUserRotation` field `usernames`: maxItems ~~`100`~~ → `2`, minItems ~~`2`~~
```
vmyroslav: Looks like redundant artifact

controllers/serviceuserrotation_controller.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (r *ServiceUserRotationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	cr := &v1alpha1.ServiceUserRotation{}
+	if err := r.Get(ctx, req.NamespacedName, cr); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	ctx = logr.NewContext(ctx, setupLogger(r.Log, cr))
+
+	if isMarkedForDeletion(cr) {
+		// Older versions added a finalizer to delete the users in Aiven.
```
vmyroslav: nit: I guess its related to the previous version of the code and can be removed.

tests/serviceuserrotation_test.go
```diff
@@ -0,0 +1,1034 @@
...
+		require.NoError(t, k8sClient.Get(s.ctx, client.ObjectKey{Namespace: s.rotation.Namespace, Name: username}, &v1alpha1.ServiceUser{}))
+		user, err := avnGen.ServiceUserGet(s.ctx, s.rotation.Spec.Project, s.rotation.Spec.ServiceName, username)
+		require.NoError(t, err)
+		require.Equal(t, username, user.Username)
+	}
+	s.waitSQL(first, "authenticated")
+	s.waitSQL(current, "authenticated")
+}
+
+func TestServiceUserRotationKafka(t *testing.T) {
```
vmyroslav: nit: we have kafka tag for this file, we can create a separate file for this test with `kafka` tag

vmyroslav (dismissed):

docs/docs/guides/serviceuser-rotation.md
```diff
@@ -0,0 +1,182 @@
...
+
+Choose exactly two distinct usernames, excluding `avnadmin`. The interval must be
+at least one hour. The user list, its order, the Secret name and the key prefix
+can't be changed after creation.
+
+Use a dedicated Secret. Let this rotation resource be the only automation that
+changes these users' passwords. Applications should use its Secret rather than
+credentials from individual `ServiceUser` resources.
+
+## How rotation works
```
staceysalamon-aiven: A small thing: I would probably put this how it works section before the instructions for setting it up.

byashimov (dismissed):
For the history: 
- I think, we could use the BE password instead of the local generated (Aiven just knows better)
- The secret is created/updated with the password before it is actually reseted on the backend, which might have watchers/triggers on that, and if the "reset" fails, then it's an issue
- I think it would be cleaner approach to use a struct to manipulate with the secret's data instead of using keys
- Keys are removed — I find it confusing to use the map as a temporary storage
- The secret is updated twice during the rotation
- The `publishedAt` is kinda tricky to read and appears both in the secret and status
- The `reconcileCredentials` called twice in the same flow — just a subtle indicator of something being a bit of complicated

Not blocking the PR, as it can be improved later.

byashimov (approved):

